PR #12, opened 2021-08-11
Update documentations for config

tzachar:
Add defaults and change wording.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,42 +53,81 @@ Configuration
 ====================
 
 The default configuration can be found in [here](./lua/cmp/config/default.lua)
+You can use your own configuration like this:
+```lua
+require'cmp'.setup {
+	completion = {
+		autocomplete = { .. },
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you very much!!!

dhruvmanila:
There are two configuration headers in the README now. I think the first one should be the installation instruction and should the configuration instructions should be removed.

tzachar:
I agree that the first one should read 'Installation Instructions'.
But why remove the second one? It makes no sense to inform users how to configure cmp.

dhruvmanila:
> But why remove the second one? It makes no sense not to inform users how to configure cmp.

Sorry for not to be clear, but I meant to remove the code block after `Then setup configuration. ...` or merge that part into the configuration section.

hrsh7th:
Ah, First one should be `Setup`

tzachar:
Want me to reedit? see #13

PR #18, opened 2021-08-12
fix: {next,prev}_item mapping

elianiva:


lua/cmp/mapping.lua
```diff
@@ -31,7 +31,7 @@ end
 mapping.next_item = function()
   return function(_, fallback)
     if vim.fn.pumvisible() == 1 then
-      vim.fn.feedkeys(vim.api.nvim_replace_termcodes('<C-n>', 'n'))
+      vim.fn.feedkeys(vim.api.nvim_replace_termcodes('<C-n>', 'n', true, true))
     else
```
hrsh7th, line 35: Sorry. I think it should be
```
vim.fn.feedkeys(vim.api.nvim_replace_termcodes('<C-n>', true, true, true), 'n'))
```

hrsh7th:
Thank you.

PR #70, opened 2021-08-25
Add mapping example to README

uga-rosa:
A mapping example that was frequently asked about has been added to README.md.
Sorry, but I don't use vsnip, so if you prefer that, please adjust accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,50 +150,20 @@ mapping = {
 ```
 
 You can specify your custom mapping function.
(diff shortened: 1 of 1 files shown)


mortezadadgar:
you could have use `cmp.utils.keymap.t` instead of defining a new function

uga-rosa:
Oh, thx!

hrsh7th:
Ah! Sorry. I  want to expose only`require('cmp')` as public API. So please don't use `reuqire('cmp.utils.*')`

hrsh7th:
LGTM! Thank you!

PR #81, opened 2021-08-26
add nofile buftype to check.ok

goolord:
per discussion in https://github.com/liuchengxu/vim-clap/issues/745, makes it so cmp won't start in buffers where `bt=nofile` (and consequently, the vim-clap window)

### DIFF
--- a/lua/cmp/utils/check.lua
+++ b/lua/cmp/utils/check.lua
@@ -2,7 +2,9 @@ local check = {}
 
 check.ok = function()
   local ng = false
-  ng = ng or vim.api.nvim_buf_get_option(0, 'buftype') == 'prompt'
+  local bt = vim.api.nvim_buf_get_option(0, 'buftype')
+  ng = ng or bt == 'prompt'
+  ng = ng or bt == 'nofile'
   ng = ng or string.sub(vim.api.nvim_get_mode().mode, 1, 1) ~= 'i'
   return not ng
 end


hrsh7th:
@goolord Sorry. Finally, I've implemented the feature that you want.
Could you try the following two ways.

1. Specify `enabled = false` for specific FileType

```vim
autocmd FileType TelescopePrompt lua require('cmp').setup.buffer { enabled = false }
```

2. Setup `enabled function` as global configuration

```lua
require('cmp')'.setup {
  enabled = function()
    return ...check some condition as you like...
  end
}
```

kunish:
> @goolord Sorry. Finally, I've implemented the feature that you want.
> Could you try the following two ways.
> 
>     1. Specify `enabled = false` for specific FileType
> 
> 
> ```viml
> autocmd FileType TelescopePrompt lua require('cmp').setup.buffer { enabled = false }
> ```
> 
>     1. Setup `enabled function` as global configuration
> 
> 
> ```lua
> require('cmp')'.setup {
>   enabled = function()
>     return ...check some condition as you like...
>   end
> }
> ```

That's a better solution in my opinion.

kunish:
> @goolord Sorry. Finally, I've implemented the feature that you want.
> Could you try the following two ways.
> 
>     1. Specify `enabled = false` for specific FileType
> 
> 
> ```viml
> autocmd FileType TelescopePrompt lua require('cmp').setup.buffer { enabled = false }
> ```
> 
>     1. Setup `enabled function` as global configuration
> 
> 
> ```lua
> require('cmp')'.setup {
>   enabled = function()
>     return ...check some condition as you like...
>   end
> }
> ```

That's a better solution in my opinion.

goolord:
ah, ok. i like the sceond option quite a bit. ty

PR #117, opened 2021-08-29
fixed supertab example

brikehn:
the current supertab example was causing autocompletion to break because `vim.api.nvim_replace_termcodes` requires 4 parameters and the example was only passing one.

Added a function to extract it and input the remaining required parameters.


hrsh7th:
Thank you.

PR #119, opened 2021-08-30
Experimental: Ghost text

hrsh7th:
This PR adds the ability to show the first candidate's preview as inline.

This is still a much experimental feature. I think we should remove this feature after VSCode's ghost text feature ships to LSP spec.

https://user-images.githubusercontent.com/629908/131284635-74da4ef7-289e-440a-abe8-bb36dc6af279.mp4


hrsh7th:
I can't think it is useful ... 🙇🏼

Issue #111, opened 2021-08-29
Lag when scrolling in a popup menu with a lot of completion items


dmitmel:
Basically what the title says. An easy way to reproduce this bug is to type `/usr/bin/` on a Linux/macOS system with the [path](https://github.com/hrsh7th/cmp-path) completion source (or with a fake source, perhaps, which just returns a bunch of results) and scroll with Up/Down arrows. I took a profile and I believe that the cause of this bug is in the `cmp.menu.get_selected_entry` function:

<https://github.com/hrsh7th/nvim-cmp/blob/24406f995ea20abba816c0356ebff1a025c18a72/lua/cmp/menu.lua#L208-L226>

This function is triggered on the `CompleteChanged` autocommand, and the call to `complete_info` has to transmit all completion items to the Lua side every time a different item is selected. Can the completion items be stored in a table on the Lua side? Alternatively, can this call be bypassed if I have `documentation` option disabled (I think the documentation float is the only user of the selected completion item, judging by the `cmp.menu.select` and `cmp.menu.unselect` functions)?

hrsh7th:
@dmitmel I've improved the implementation. Could you test with latest ?

dmitmel:
No, it doesn't, the slow path is still taken... Isn't `v:complete_item` intended for use in `CompleteDone`? By the way, here's a source that can be used to diagnose the problem:

```lua
local source = {}
source.new = function()
  return setmetatable({}, { __index = source })
end
function source:is_available()
  return true
end
function source:get_keyword_pattern(params)
  return '.+'
end
function source:complete(params, callback)
  local items = {}
  for i = 1, 10000 do
    items[i] = { label = string.format('%d', i) }
  end
  callback(items)
end
cmp.register_source('lag_machine', source.new())
```

dmitmel:
I think the following patch should do the job:

```diff
diff --git a/lua/cmp/menu.lua b/lua/cmp/menu.lua
index cd24b11..6c353af 100644
--- a/lua/cmp/menu.lua
+++ b/lua/cmp/menu.lua
@@ -217,9 +217,8 @@ menu.get_selected_entry = function(self)
   if selected == -1 then
     return nil
   end
-  local items = vim.fn.complete_info({ 'items' }).items
 
-  local completed_item = items[math.max(selected, 0) + 1] or {}
+  local completed_item = self.items[math.max(selected, 0) + 1] or {}
   if not completed_item.user_data then
     return nil
   end
[...]

PR #147, opened 2021-09-03
Bring back the duplicates filter due to shifting completion item indexes after deduplication (continuation of fixes for #111)

dmitmel:
Alright alright, I realize that you are probably tired of #111 already, but I have found a pretty serious bug related to deduplication, which you removed in commit e599f0825cc6de637a70aee9b941698417d27768. As an example, let's say that the following completion list has to be displayed:

```lua
{
  {
    abbr = "print(...)~",
    dup = 1,
    empty = 1,
    equal = 1,
    kind = "Snippet",
    user_data = { cmp = 1911 },
    word = "print"
  }, {
    abbr = "print",
    dup = 0,
    empty = 1,
    equal = 1,
    kind = "Text",
    user_data = { cmp = 1709 },
    word = "print"
  }, {
    abbr = "printf(...)~",
    dup = 1,
    empty = 1,
    equal = 1,
    kind = "Snippet",
    user_data = { cmp = 1912 },
    word = "printf"
  }, {
    abbr = "vim_print",
    dup = 0,
    empty = 1,
    equal = 1,
    kind = "Text",
    user_data = { cmp = 1126 },
    word = "vim_print"
  }
}
```

The first three entries come from the Lua language server (`sumneko_lua`), the last one is from the cmp-buffer source. Now, after nvim-cmp calls `vim.fn.complete()` which performs deduplication internally, only `print(...)~`, `printf(...)~` and `vim_print` will be displayed to the user. However, when I select `vim_print`, its index (returned by `vim.fn.complete_info().selected`) will be 2 and not 3, but nvim-cmp will access the entry with index 2, will get the `printf` one and try to expand it instead of the `vim_print` which I have requested.

So basically, as a solution we must deduplicate on the Lua side, store the results in a table and then we will be safe to access that table with indices returned by `vim.fn.complete_info().selected`. I re-used most of the code from the commit e599f0825cc6de637a70aee9b941698417d27768, but changed the logic to use `item.word` instead of `item.abbr` because that's what Vim does. Also, as a sidenote, our implementation might very well be faster on long lists than Neovim's ([code](https://github.com/neovim/neovim/blob/5f8518b3f04c2295be24912dba78d3e34ca74b91/src/nvim/edit.c#L2352-L2366)) because they use linear lookup and we have a hashtable!

### DIFF
--- a/lua/cmp/menu.lua
+++ b/lua/cmp/menu.lua
@@ -16,6 +16,7 @@ local check = require('cmp.utils.check')
 ---@field public on_select fun(e: cmp.Entry)
 ---@field public items vim.CompletedItem[]
 ---@field public entries cmp.Entry[]
+---@field public deduped_entries cmp.Entry[]
 ---@field public selected_entry cmp.Entry|nil
 ---@field public context cmp.Context
 ---@field public resolve_dedup fun(callback: function)
@@ -58,6 +59,7 @@ menu.reset = function(self)
   self.offset = nil
   self.items = {}
   self.entries = {}
+  self.deduped_entries = {}
   self.context = nil
   self.preselect = 0
   self:close()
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM 👍🏼

Issue #165, opened 2021-09-06
attempt to index local 'c'


gegoune:
Started getting

```
Error executing vim.schedule lua callback: Vim(lua):E5108: Error executing lua ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:62: attempt to index local 'c' (a boolean value)
Error detected while processing CompleteChanged Autocommands for "*":
E5108: Error executing lua ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:62: attempt to index local 'c' (a boolean value)
```

when typing anything and when completion menu pops up (have automatic completion set up).

Fixable by disabling 

```lua
experimental.ghost_text = true
```

and started appearing after https://github.com/hrsh7th/nvim-cmp/commit/b6b15d5f6e46643462b5e62269e7babdab17331c got merged.

gegoune:
Setting `hl_group` as per #166 fixes the issue.

hrsh7th:
I've fixed it.

rohit-kumar-j:
@hrsh7th, currently the tested commit (origin/main) [`935b406`](https://github.com/hrsh7th/nvim-cmp/commit/935b4069ce73b60ba9075bf05ee6ab50ed3af1a9) errors like this:
```bash
cmp:GHOST_TEXT: Error executing lua: .../nvim-data/lazy/nvim-cmp/lua/cmp/view/ghost_text_view.lua:40: attempt to index local 'c' (a boolean value)
```
Switching back to 'v0.0.1' fixes this.

PR #166, opened 2021-09-06
[fix, docs]: add default ghost text hl group, update README

McAuleyPenney:
Edit: Thank you for adding the hl group for ghost text! I added a default so that specifying the hl group is not absolutely necessary and added a little piece of documentation about the ghost text hl group in the README. If this isn't desired, I totally understand!

Why:
    - using ghost text without specifying hl group gave errors
    - the ghost text hl group has no docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,14 @@ Specify whether to display ghost text.
 Default: `false`
 
 
+#### experimental.ghost_text.hl_group (type: string)
+
+Specify the hightlight group used for ghost text.
+
+Default: `Comment`
+
+
+
 Programatic API
 ====================
 
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -36,8 +36,10 @@ core.menu = menu.new({
 core.ghost_text = function(e)
   vim.api.nvim_buf_clear_namespace(0, core.GHOST_TEXT_NS, 0, -1)
 
-  local c = config.get().experimental.ghost_text
-  if not c then
+  local experimental_cfg = config.get().experimental
+
+  local use_ghost_text = experimental_cfg.ghost_text
+  if not use_ghost_text then
     return
   end
 
@@ -59,9 +61,10 @@ core.ghost_text = function(e)
   if #text > 0 then
     vim.api.nvim_buf_set_extmark(ctx.bufnr, core.GHOST_TEXT_NS, ctx.cursor.row - 1, ctx.cursor.col - 1, {
       right_gravity = false,
-      virt_text = { { text, c.hl_group or "Comment" } },
+      virt_text = { { text, experimental_cfg.ghost_text_highlight_group } },
       virt_text_pos = 'overlay',
       virt_text_win_col = ctx.virtcol - 1,
+      hl_mode = 'combine',
       priority = 1,
     })
   end
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -59,7 +59,7 @@ core.ghost_text = function(e)
   if #text > 0 then
     vim.api.nvim_buf_set_extmark(ctx.bufnr, core.GHOST_TEXT_NS, ctx.cursor.row - 1, ctx.cursor.col - 1, {
       right_gravity = false,
-      virt_text = { { text, c.hl_group } },
+      virt_text = { { text, c.hl_group or "Comment" } },
       virt_text_pos = 'overlay',
       virt_text_win_col = ctx.virtcol - 1,
       priority = 1,
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -66,6 +66,7 @@ return function()
 
     experimental = {
       ghost_text = false,
+      ghost_text_highlight_group = "Comment",
     },
 
     sources = {},


gegoune:
Fixes #165

README.md
```diff
@@ -385,6 +385,14 @@ Specify whether to display ghost text.
 Default: `false`
```
hrsh7th, line 385: Thank you for your contribution.
Could you modify `experimental.ghost_text` type as `type: cmp.GhostTextConfig|"false"` too?

McAuleyPenney: I'd be happy to make a change! I'm learning something new here though. What does this change do and where does it go?

hrsh7th: All configuration options were defined here https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/types/cmp.lua#L93

I'd forgotten to follow it with README.md

McAuleyPenney: Like this? This is my first time working with Lua annotations 😓  
image

haze:
Just so there aren't two competing PRs:

@McAuleyPenney there is a bug that prevents user defined highlight groups from being used.

Line `39` of `core.lua` assigns `c` to `config.get().experimental.ghost_text`, while it should just be `config.get().experimental`. This means that any attempt to grab `hl_group` will fail because `ghost_text` has no property. I'd also suggest renaming `hl_group` to `ghost_text_highlight_group`. (cc: @hrsh7th)

McAuleyPenney:
@haze I am taking a look at this now. 

1. I've changed the `hl_group` variable name     ✅ 

2. I gave `buf_set_extmark` `hl_mode = 'combine'`    ✅ 

3.  My configuration, using the fork I created, is currently allowing me to choose my ghost text highlight, but the config for ghost text is forced  into being a table. I was trying to figure out why earlier today and you solved it.

With my configuration and fork as they are, using a custom hightlight I receive:
image

and with the highlight config commented out, I receive the "Comment" hl group:
image

but, if the user wants to simply use the default by saying `ghost_text = true`, then I receive this error:
image

The modification to `config.get().experimental.ghost_text` fixes that, so this is the consequent appropriate config:
image

I will push these changes to my fork right now

haze:
I think the `or` is okay. I think retaining the old functionality for users who blindly update their plugins is desired (while also expanding it, of course). Then again, this isn't my plugin, that decision is left up to hrsh7th.

gegoune:
From user's perspective it made more sense when it was

```lua
experimental = {
  ghost_text = { hl_group = 'Group' }
}
```

Or if no more options are ever expected it could be simplified to

```lua
experimental = {
  ghost_text = true
}
```
using some default or

```lua
experimental = {
  ghost_text = 'Group'
}
```

haze:
@gegoune That may also be a desired approach, just assumed that the options would reside in a flat namespace until they graduated from experimental status

gegoune:
> I think retaining the old functionality for users who blindly update their plugins is desired

Quote the opposite, as long as feature is experimental it is best time to change things without worrying about users at all.

haze:
I only said that because it happened to me, all of a sudden ghost text looked ugly and thats how I came upon the new option and bug. Maybe you're right

McAuleyPenney:
> > I think retaining the old functionality for users who blindly update their plugins is desired
> 
> Quote the opposite, as long as feature is experimental it is best time to change things without worrying about users at all.

Unless I am mistaken, moving it from where I had it before to where ( and how ) it is now isn't a functional change. This just allows for it to be explicit in the default config, which will take precedence assuming the user doesn't define an option, right? That way, the default isn't buried in the source. Again, totally possible that I'm wrong.

gegoune:
> @gegoune That may also be a desired approach, just assumed that the options would reside in a flat namespace until they graduated from experimental status

Ah, not bad assumption actually. @hrsh7th ideas for the feature and its configurability will dictate what it will look like once out of experimental, but I would guess `experimental` is just a wrapper for 'expect breaking changes with no notice and unexpected behaviour' but once feature matures its key would just pop out to top level (or nested under some other key) but with already tested implementation and configuration.

McAuleyPenney:
> From user's perspective it made more sense when it was
> 
> ```lua
> experimental = {
>   ghost_text = { hl_group = 'Group' }
> }
> ```
> 
> Or if no more options are ever expected it could be simplified to
> 
> ```lua
> experimental = {
>   ghost_text = true
> }
> ```
> 
> using some default or
> 
> ```lua
> experimental = {
>   ghost_text = 'Group'
> }
> ```

The second way that you list here was actually giving me errors when I tested it with my original pull request. I think I have a screenshot of it in the post I made above. I don't know the source code well enough to solve it, unfortunately. With how I had it, to use ghost text with the default, you had to use an empty table. I think I also had a screenshot of that

gegoune:
It would require checking type of key's value and returning some default if it's true, if it's string just use group from the string.

haze:
Any of the solutions that gegoune suggested would work, you would just need to clarify where the highlight group comes from. In the first example, it would be `experimental.ghost_text` and if that object isn't nil, then the property `hl_group` on that table would be the group. In the second, there is no group defined so it would exist as another variable in the `experimental` table. In the last, (which would probably incur the biggest change) would be to assume that if the `ghost_text` property on the `experimental` table is defined, to use the **value** of that key as the highlight group. If there is no `ghost_text` property then it doesn't matter if the value is nil since there is no invocation that requires a highlight group

McAuleyPenney:
Alright, I familiarized myself and made some changes. See the screenshots for a display of the functionality

```lua
core.ghost_text = function(e)
  vim.api.nvim_buf_clear_namespace(0, core.GHOST_TEXT_NS, 0, -1)

  local experimental_cfg = config.get().experimental

  local use_ghost_text = experimental_cfg.ghost_text
  if not use_ghost_text then
    return
  end

  if not e then
    return
  end

  local ctx = context.new()
  if ctx.cursor_after_line ~= '' then
    return
  end

  local diff = ctx.cursor.col - e:get_offset()
  local text = e:get_insert_text()
  if e.completion_item.insertTextFormat == types.lsp.InsertTextFormat.Snippet then
    text = vim.lsp.util.parse_snippet(text)
  end
  text = string.sub(str.oneline(text), diff + 1)
  if #text > 0 then
    vim.api.nvim_buf_set_extmark(ctx.bufnr, core.GHOST_TEXT_NS, ctx.cursor.row - 1, ctx.cursor.col - 1, {
      right_gravity = false,
      virt_text = { { text, experimental_cfg.ghost_text_highlight_group } },
      virt_text_pos = 'overlay',
      virt_text_win_col = ctx.virtcol - 1,
      hl_mode = 'combine',
      priority = 1,
    })
  end
end 
```
No ghost text:
image

default highlight:
image

custom highlight:
image

haze:
I'd probably just leave that up to hrsh7th when the feature stabilizes or something, it looks okay to me as is

McAuleyPenney:
For sure. I'm going to push this and abdicate control over to hrsh7th

lua/cmp/core.lua
```diff
@@ -59,9 +61,10 @@ core.ghost_text = function(e)
   if #text > 0 then
     vim.api.nvim_buf_set_extmark(ctx.bufnr, core.GHOST_TEXT_NS, ctx.cursor.row - 1, ctx.cursor.col - 1, {
       right_gravity = false,
-      virt_text = { { text, c.hl_group } },
+      virt_text = { { text, experimental_cfg.ghost_text_highlight_group } },
       virt_text_pos = 'overlay',
       virt_text_win_col = ctx.virtcol - 1,
+      hl_mode = 'combine',
```
hrsh7th, line 67: hl_mode combine is good.

hrsh7th:
Sorry. I don't prefer the name of `ghost_text_highlight_group`.
I choose the `hl_mode` (It is the same name of extmark API)

hrsh7th:
Please don't modify the API signature. `{ experimental = { ghost_text = { hl_group = 'Comment' } } }`

McAuleyPenney:
Okay, changes removed except `hl_mode = combine`👍🏻 

Was the change I made to the types file acceptable? If so, I'll push this and we can get this closed

haze:
I don't believe that PR fixes the bug where the wrong variable is referenced, it will always default to the `Comment` highlight group

hrsh7th:
sorry. I am happy with your contribution!
If you have any other improvements, please feel free to PR!

McAuleyPenney:
> sorry. I am happy with your contribution!
> 
> If you have any other improvements, please feel free to PR!

No worries! I'm not bothered by it, I'm glad that the change got implemented! I appreciate your work!

PR #168, opened 2021-09-06
Reduce use of VimL

Iron-E:
Hello,

Thank you for all your work on this plugin! I noticed while reading through the code that there are many instances where `vim.cmd` and/or `vim.fn` are used when there are `vim.api` functions that could be used instead. The motivation for this is that `vim.api` functions take anywhere from ¹/₂ to ²/₃ the time to complete as VimL, so using them consistently will improve the speed of the plugin overall.

I have gone through already and replaced the functions with API equivalents (except for in the spec, where I figure performance doesn't _really_ matter… feel free to correct me). All I need to do is test it to make sure there are no breakages.

Let me know if there are any questions about what I've done. I will be happy to answer any questions.


lua/cmp/utils/spec.lua
```diff
@@ -5,16 +5,14 @@ local types = require('cmp.types')
...
 
 spec.before = function()
-  vim.cmd [[
-    bdelete!
-    enew!
-    setlocal virtualedit=all
-  ]]
+	vim.api.nvim_buf_delete(0, {force = true})
+	vim.api.nvim_command [[enew!]]
+	vim.opt.virtualedit = 'all'
```
Iron-E, line 10: Just to explain this… `virtualedit` has no local option. From `:setlocal`:

> Not all options have a local value.  If the option does not have a local	value the global value is set.

Therefore `:setlocal` was equivalent to `:set` here. Trying to `:setlocal` with Lua resulted in an error, which confirms it. So `vim.opt` will have to do.

lua/cmp/float.lua
```diff
@@ -123,7 +123,7 @@ float.scroll = function(self, delta)
 
     vim.defer_fn(function()
       vim.api.nvim_buf_call(buf, function()
-        vim.cmd('normal! ' .. top .. 'zt')
+        vim.api.nvim_command('normal! ' .. top .. 'zt')
```
Iron-E, line 126: Here, `vim.api.nvim_command` is faster than `vim.cmd` because `vim.cmd` takes multiple lines into account whereas `vim.api.nvim_command` only works on one. Thus the logic of dealing with multiple lines is absent from `vim.api.nvim_command`. 

You can also personally verify this by running `vim.loop.hrtime` on both statements.

lua/cmp/context.lua
```diff
@@ -55,8 +55,9 @@ context.new = function(prev_context, option)
   self.cursor_line = vim.api.nvim_get_current_line()
   self.virtcol = vim.fn.virtcol('.')
   self.cursor = {}
-  self.cursor.row = vim.api.nvim_win_get_cursor(0)[1]
-  self.cursor.col = vim.api.nvim_win_get_cursor(0)[2] + 1
+  local cursor = vim.api.nvim_win_get_cursor(0)
```
Iron-E, line 58: Here, calling `nvim_win_get_cursor`once is more efficient than calling twice. Otherwise I made no change.

lua/cmp/utils/spec.lua
yujinyuz: what about `vim.opt_local.virtualedit`?

Iron-E: I tried that along with `bo` and `wo`. The problem is that `virtualedit` is a global option (see `:h 'virtualedit'`), so even doing `:setlocal` has no specific effect and just resolves to a `:set`.

hrsh7th:
Thank you very much. I'll check it.

hrsh7th:
LGTM. Could you solve the conflicts?

Iron-E:
I rebased on main, should be good now.

Iron-E:
I rebased on main, should be good now.

README.md
```diff
@@ -174,18 +174,18 @@ You can specify your own custom mapping function.
 
 ```lua
 local check_back_space = function()
```
gegoune, line 176: This isn't strictly related to this PR so apologies for noise, but I am trying to figure out at the moment what is this function for and how having this conditional in tab mapping changes things. Would appreciate if someone could shed some light here.

Iron-E: It determines whether the cursor is within/in-front-of a word (returning `false`), or there is whitespace behind the caret (returning `true`). Based on this, a `<Tab>` mapping can determine whether it should indent or begin completion (whether to cycle completions is based on `pumvisible`, the first check). After all, without a word to complete we want to allow the `<Tab>` to passthrough or else we cannot indent anymore.

See the example in the README or my personal config, knowing the above should hopefully be enough to walk through line-by-line.

gegoune: Ah, so I was pretty close thinking about it. This then means that if I don't have tab set up to trigger completion in the first place this isn't really relevant to me at all, is it?

And great thanks for that explanation! I have seen this function around in compe as well, for quite a while and was always bit surprised that with or without it there wasn't any difference I could see.

Iron-E: > This then means that if I don't have tab set up to trigger completion in the first place this isn't really relevant to me at all, is it?

Nope! If you're not using `<Tab>`, it won't matter.

> I have seen this function around in compe as well, for quite a while

The first time I saw it was [back in `coc.nvim`](https://github.com/neoclide/coc.nvim#example-vim-configuration). Blame shows [this commit](https://github.com/neoclide/coc.nvim/commit/8b55463159de9020bf61f8f0074ae7bad591d4de), though in that commit the function already existed. So, who knows how old it is or where it really comes from 🤔

Happy to help!

gegoune: Awesome, great thanks! You have helped me a lot here now!

hrsh7th:
Thank you!

PR #195, opened 2021-09-13
Slight complexity reduction

tzachar:
Do not loop twice over sources table, instead keep a table mapping names
to sources

### DIFF
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -72,6 +72,9 @@ end
 ---@type table<number, cmp.Source>
 core.sources = {}
 
+---@type table<string, cmp.Source>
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you. It is a improvement.

hrsh7th:
LGTM! Thank you!

hrsh7th:
Ah. Sorry! It has a side-effect. I will revert it.

tzachar:
What is the side effect? Can it be mitigated?

hrsh7th:
For example, `nvim_lsp` named source can be multiple. (For example, css language server & emmet-ls.)
But this implementation will group the sources by name.

PR #196, opened 2021-09-14
Rework #195

tzachar:
Change sources table to hold a table mapping source name to a table of sources.
Solves Issues raised in #195

### DIFF
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -72,7 +72,7 @@ end
 ---@type table<number, cmp.Source>
 core.sources = {}
 
----@type table<string, table<cmp.Source>>
+---@type table<string, cmp.Source[]>
 core.sources_by_name = {}
 
 ---@type cmp.Context
@@ -82,7 +82,7 @@ core.context = context.new()
 ---@param s cmp.Source
 core.register_source = function(s)
   core.sources[s.id] = s
-  if not vim.tbl_contains(core.sources_by_name, s.name) then
+  if not core.sources_by_name[s.name] then
     core.sources_by_name[s.name] = {}
   end
   table.insert(core.sources_by_name[s.name], s)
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM.

lua/cmp/core.lua
```diff
@@ -72,18 +72,30 @@ end
 ---@type table<number, cmp.Source>
 core.sources = {}
 
+---@type table<string, table<cmp.Source>>
```
hrsh7th, line 75: I prefer to annotate by `cmp.Source[]` instead.

lua/cmp/core.lua
```diff
@@ -72,18 +72,30 @@ end
...
+core.sources_by_name = {}
+
 ---@type cmp.Context
 core.context = context.new()
 
 ---Register source
 ---@param s cmp.Source
 core.register_source = function(s)
   core.sources[s.id] = s
+  if not vim.tbl_contains(core.sources_by_name, s.name) then
```
hrsh7th, line 85: You should use `if not core.sources_by_name[s.name] then` instead. Because it is not a `List` but `Dict`.

lua/cmp/core.lua
```diff
@@ -118,12 +130,10 @@ end
 core.get_sources = function(statuses)
   local sources = {}
   for _, c in pairs(config.get().sources) do
-    for _, s in pairs(core.sources) do
-      if c.name == s.name then
-        if not statuses or vim.tbl_contains(statuses, s.status) then
-          if s:is_available() then
-            table.insert(sources, s)
-          end
+    for _, s in ipairs(core.sources_by_name[c.name] or {}) do
```
hrsh7th, line 133: LGTM.

lua/cmp/core.lua
tzachar: Not sure what you mean.

tzachar: ok got it...

hrsh7th: You should use `---@type table<string cmp.Source[]>` instead.

PR #198, opened 2021-09-15
Experimental: Active preview

amazingefren:
This PR adds the ability to show the first candidate's preview as ~~inline~~ if it were selected.

This is still a much experimental feature. I think we should remove this feature after VSCode's ~~ghost text~~ active preview feature ships to LSP spec.

Honestly though, this is a pretty useful feature to have when using `noselect` in completeopts

Note: this method works without interfering with the get_selected_entry call

https://user-images.githubusercontent.com/48777455/133379793-cc80637c-a4e4-424b-919b-16b4e77e1fd6.mp4

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -19,7 +19,6 @@ return function()
       get_trigger_characters = function(trigger_characters)
         return trigger_characters
       end,
-      prepreview = false,
     },
 
     snippet = {
@@ -70,6 +69,7 @@ return function()
 
     experimental = {
       ghost_text = false,
+      active_preview = false,
     },
 
     sources = {},
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -19,6 +19,7 @@ return function()
       get_trigger_characters = function(trigger_characters)
         return trigger_characters
       end,
+      prepreview = false,
     },
 
     snippet = {
--- a/lua/cmp/menu.lua
+++ b/lua/cmp/menu.lua
@@ -32,6 +32,9 @@ menu.new = function(opts)
   self:reset()
   autocmd.subscribe('CompleteChanged', function()
     local e = self:get_selected_entry()
+    if not e and config.get().completion.prepreview then
+      e = self:get_first_entry()
+    end
     if e then
       self:select(e)
     else
@@ -40,6 +43,7 @@ menu.new = function(opts)
   end)
   return self
 end
+
 
 ---Close menu
 menu.close = function(self)
@@ -180,6 +184,10 @@ menu.select = function(self, e)
   e:resolve(self.resolve_dedup(vim.schedule_wrap(function()
     if self:get_selected_entry() == e then
       self.float:show(e)
+    elseif config.get().completion.prepreview then
+      if self:get_first_entry() == e then
+        self.float:show(e)
+      end
     end
   end)))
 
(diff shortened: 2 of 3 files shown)


hrsh7th:
Isn't https://github.com/hrsh7th/nvim-cmp#how-to-set-up-like-nvim-compes-preselect--always enough?

amazingefren:
That is what I was using, however, 
assume mapping this
```
["<CR>"] = cmp.mapping.confirm({
  behavior = cmp.ConfirmBehavior.Replace,
  select = false
})
```
without "noselect", it will choose to complete the selected item 100% of the time.
while this is a niche use-case, this method provides functionality to support that map

to extend to that,
assume mapping this
```
["<BS>"] = cmp.abort()
```
without "noselect", it will simply attempt to abort everytime pumvisible == 1

side-note: cmp.abort()
changing
```
cmp.abort = function () 
  - if vim.fn.pumvisible() == 1
  + if vim.fn.pumvisbile() == 1 and ${complete_info.selection} >= 0
    ...
```
 provides full functionalitity to the <BS> abort without interfering when "noselect" is not set

overall, with my commited changes + the above side note, it should allow for a workflow in which you can use BS to cancel a selection (when above -1) && use CR to go to new line instead of autocompleting the first line && activate snippet completion with CR when selected

(while providing a preview)

hrsh7th:
```lua
['<BS>'] = function(fallback)
  if vim.fn.pumvisible() == 1 then
    return cmp.close()
  end
  fallback()
end
```

nvim-cmp's mapping feature is fully customizable (but it's difficult a bit).
You can customize <BS> behavior by the above setting.

amazingefren:
I was using a custom map that had 
```
if vim.fn.pumvisible() == 1 and vim.fn.complete_info({'selected'}).selected > -1
  cmp.abort()
else
cmp.utils.keymap.feedkeys(cmp_keymap.t('<BS>'),'n',cmp.core.reset())
```

now assume you wanted a preview using `noinsert`
if I select item 5 from completion menu, and hit BS, boom it aborted and good to go
if I select item 0 from completion menu, and hit BS, boom it worked again
if I select item -1 from completion menu, and hit BS, it aborted (even though selected was -1)

now imagine hitting BS to delete a character, delete, then cmp decides to open the completion menu
without `noselect` 'complete_info.selected' says its 0, so instead of forwarding the <BS> key, instead
cmp will interpret that selected is 0, so it aborts instead, therefore requiring 2 BS inputs for each character 
to be deleted every time cmp decides to pop up

*this clause is for the side-note in my previous comment*

all I tried to implement was a way to have cmp ACT like it had preselect without vim reporting that the user had selected something greater than -1, provides a seamless experience for users preferring to SEE the preview window, without worrying that hitting the CR / BS key being interpreted incorrectly as completion actions

hrsh7th:
OK. Probably, I've understood.

1. You want to show `docs window` even if the item wasn't selected (in this case, the first item will be chosen as doc window)
2. But the setting like `preselect="always"` has a side-effect of `<CR>` so can't usable.

Thank you for understanding and using nvim-cmp so far.
However, this PR is unacceptable.

In my opinion, it will be good to add "confirmation.get_target_item" and "documentation.get_target_item" as configurable options.

amazingefren:
no problem, I love the plugin

I didn't fully plan this PR to be a direct implementation, simply just a quick/rough concept for a feature that might benefit a small amount of users who prefer not to use completion without `noselect` and use a more natural feeling implementation of `<CR>` and `<BS>`

just so I'm clear,
is this a feature you WOULD like to see implemented?

hrsh7th:
I think this PR's configuration wouldn't like to implement to nvim-cmp.

But I think your opinion can be understandable so I suggested `In my opinion, it will be good to add "documentation.get_target_item" as configurable options.` for your expectation.

amazingefren:
gotcha,
feel free to close this PR as you see fit, I personally will not be taking up the implementation 

ty :)

hrsh7th:
Thank you~! I'm closing now.

PR #199, opened 2021-09-15
Fix typos in preselect options in README

minhduc0711:


hrsh7th:
You are correct. Thank you!

PR #224, opened 2021-09-19
WIP: Use floating window for completion menus

hrsh7th:
I don't think it can be completed... (and it will introduce a lot of breaking change.)


hrsh7th:
Anyway, It's mostly working.

hrsh7th:
I don't know the reason but it's improve performance 😂

hrsh7th:
Support scrollbar.

https://user-images.githubusercontent.com/629908/134109000-aab23f4b-a1d1-49eb-8caf-035d095d784e.mp4

uga-rosa:
I get an error or the cursor jumps to the beginning of the column. When I cut it down to this level, it worked fine.

```vim
call plug#begin("/tmp/cmp-test")
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
call plug#end()

set completeopt=menu,menuone,noselect

lua <<EOF
local cmp = require'cmp'

cmp.setup({
  mapping = {
    ['<C-d>'] = cmp.mapping.scroll_docs(-4),
    ['<C-f>'] = cmp.mapping.scroll_docs(4),
    ['<C-Space>'] = cmp.mapping.complete(),
    ['<C-e>'] = cmp.mapping.close(),
  },
  sources = {
    { name = 'buffer' },
  }
})
EOF
```

hrsh7th:
I've met the same problem. I think currently the problem should be solved.
Of course, It can be another root cause...

uga-rosa:
I found the cause, it seems to behave strangely when completeopt is not set.

uga-rosa:
This is a different problem, but the menu only appears at the bottom, so if the cursor is at the bottom, the menu is not visible.

uga-rosa:
Cursor jumping has been fixed!!

hrsh7th:
> This is a different problem, but the menu only appears at the bottom, so if the cursor is at the bottom, the menu is not visible.
Currently, this is intended but hm... I will implement it.

Shougo:
@hrsh7th Hi.  I want to test the feature.  What is the fancy menu config?

ericpubu:
I've been using this branch today, and it works mostly fine!
this is a really nice change, thanks!

The only thing I've found is that when I enter through the completion menu and out (`select_next_item` and `select_prev_item`), it's inserting the value, making the undo behaves weird, on the main branch just one undo removes the last value:

https://user-images.githubusercontent.com/18682359/134386797-aec1146d-17ba-42b8-86c7-9b70adc0e4f1.mov

Maybe you have already found this, but I'll leave it here if it's useful! ignore it otherwise.

tzachar:
This looks really good.
Can we make the color scheme configurable?

hrsh7th:
@tzachar I've implemented customi highlights.

You can look `plugin/cmp.lua` for it.

tzachar:
> @tzachar I've implemented customi highlights.
> 
> You can look `plugin/cmp.lua` for it.

Cool :)

rhcher:
Hi, I've been using this branch for a while now. So far it is working well. The only insignificant glitch is in the documentation for entries.
This is in the master branch:
image
And this is in the custom-menu branch:
image
If you are already aware of this problem, please ignore it.

uga-rosa:
image
I noticed something strange. When a non-consecutive match occurs, the latter part of the match is italicized. This is independent of the color scheme.

akinsho:
@uga-rosa I believe this is intentional, there is a special highlight for characters that were matched fuzzily `CmpItemAbbrMatchFuzzy` which seems to have a default italic highlight which I'm sure you can override if you wanted 🤷🏿‍♂️

uga-rosa:
@akinsho I see. I didn't read the readme of this branch properly:(
I personally don't like italics because they can break the display.

tzachar:
Been using this branch for a while now.
Some issues:

1) When not mapping <C-c> to close the completion window, and pressing '<C-c>' while the completion window is open, this is what I get:
bad_C-c

2) When I map <C-e> like this:

```lua
['<C-e>'] = cmp.mapping.abort()
```

or this:
```lua
['<C-e>'] = cmp.mapping.close()
```
Pressing <C-e> repeatedly clones the following line:


bad_c_e

uga-rosa:
@tzachar The first is probably due to the fact that `insertleave` is not fired, which is a specification of `i_CTRL-c`.
The second is the original `i_CTRL-e` behavior. I think this was intended. When the cmp function does not fire, the original map should be used.

hrsh7th:
I've added todos

tzachar:
@hrsh7th I now understand why `<C-e>` behaves as such... This is basically the default mapping of `<C-e>`... The question is why the fallback is called in this case.

uga-rosa:
https://github.com/hrsh7th/nvim-cmp/blob/ab74c9a7eb1f1a23ec4dde9354965fdd51a6f598/lua/cmp/init.lua#L56

I think it's here. `feedkey` that was originally included to close menu is no longer needed.

hrsh7th:
@uga-rosa Yes. I think `cmp.mapping.{close,abort}` should call `view:{close,abort}()`.
Current implementation is depending on native menu.

uga-rosa:
@hrsh7th Since `view:close()` is called in `cmp.core:reset()`, which is called one line above this one, I think it will work if you just delete this line.

hrsh7th:
Hm... indeed. But we need the `<C-e>` to clear `v:completed_item` after completion...
I will remove `<C-e>` for now.

hrsh7th:
vim's native completion menu will save state to `v:completed_item`. It isn't cleared when closed the pum.
If `v:completed_item` exists, the `<C-n>` does not work...

tzachar:
Im not sure `<C-e>` is the only problem. I think this is something more systematic. When I map `<C-c>` to `cmp.mapping.abort`, 
Im getting a capital `c` b4 the menu is closed.

cmp should not pass on the keypress if the menu is open.

uga-rosa:
Hmm? In my environment, by deleting line 56, `<C-e>` works now correctly, regardless of experimental.native_menu.

hrsh7th:
I think `cmp.mapping.abort` and `cmp.mapping.close` works propely now.

tzachar:
Works like a charm :)

hrsh7th:
I'm considering how to specify the border and winhighlight option by the user...

https://user-images.githubusercontent.com/629908/134810491-a7eec8f8-94ae-4e76-96fe-37525f5b3b35.mp4

hood:
A simple question: will this feature be optional?

hrsh7th:
This feature is the default.
It can be disabled with `cmp.setup {experimental = {native_menu = true}}`.

I would like to have only a custom menu in the future.

akinsho:
> I'm considering how to specify the border and winhighlight option by the user

@hrsh7th why not take the border in the exact format compatible with `nvim_open_win` but winhiglight can be a list which is *maybe* easier to follow than `NormalFloat:WarningMsg,FloatBorder:ErrorMsg` e.g.
```lua
border = 'single' | 'double' | 'rounded' | {"/", "-", "\\", "|"}
winhighlight = {{'FloatNormal', 'WarningMsg'}, {'FloatBorder', 'ErrorMsg'}}
```
Just a random suggestion, feel free to ignore 🤷🏿‍♂️ 

Also I noticed that the scroll bar thumb is customisable I think but not the track, it's just the same color as the background whereas other custom float scrollbars allow styling each bit for example `wilder.nvim`'s looks like

image

hrsh7th:
@akinsho Looks very good! But is super difficult...

The reason is that default appearance uses border (whitespace) too.
So we can't determine should shift the scrollbar or not.

We should introduce `win_mode` option. If win_mode is specified as `single/double/rounded`, we should shift the scrollbar to the window inner.

akinsho:
@hrsh7th it's definitely a *super minor* point I thought maybe it was already possible but the default highlighting was just set to the same colour as the background and it would just be a matter of changing the highlight group. If it's at all complex, then just disregard the suggestion

hood:
> This feature is the default.
> It can be disabled with `cmp.setup {experimental = {native_menu = true}}`.
> 
> I would like to have only a custom menu in the future.

IMHO keeping the faculty to choose between the two options could prove a better move. This way if futures users'll want to use the OG Pmenu for any reason they'll be able keep using `nvim-cmp` instead of having to find alternatives just for a matter of aesthetics. Just my 2cents ofc, great job for all the work anyway.

McAuleyPenney:
Hey, the custom menu is super cool, thank you for this. I'm trying it out right now and one thing I notice is that, when ghost text is also being used, the ghost text is left even after backspacing the current word. 

For example, I am typing the word "for" in this first screenshot:
image


 After backspacing through it:
image

I recognize that this is not a priority at the moment but figured that it might be worth mentioning.

hrsh7th:
I will merge after a while.

dmitmel:
> > [...]
> > I would like to have only a custom menu in the future.
> 
> IMHO keeping the faculty to choose between the two options could prove a better move. This way if futures users'll want to use the OG Pmenu for any reason [...]

One reason may be that the built-in completion menu has better integration with Neovim GUIs. I too think that keeping this as an option is the right way.

hrsh7th:
Remaining enhancements.

1. Some window mode support
- Only need to support `rounded / single / double / default`.
- The `default` means to add padding on the left / right side.
- Special support for scrollbar position in bordered mode.

2. Solve matched character highlights with caches
- It's a bit complicated...

Shatur:
> I understand that I should keep the native menu mode, but then I shouldn't have this feature. I also think due to maintenance costs.

If the new menu is faster and nicer - I would okay to use the only new menu.

hrsh7th:
My plan is
1. First, we support both native and custom to avoid community confusion.
2. Wait and see for a moment then we decide the native menu can be removed or not. (I hope it can be removed ...)

gegoune:
> Only need to support rounded / single / double / default.

Why not allow passing `border` same way as for `nvim_open_win`? It's very flexible.

gegoune:
It's all nitpicking, but why does it have to be out of the window? Shouldn't it actually be within the border of the window?

hrsh7th:
#### The visible border mode.
We should locate the scrollbar to -2 offset from the right edge.

#### The invisible border mode.
We should locate the scrollbar to -1 or 0 offset from the right edge.

sQVe:
I ran this branch when working today and really enjoyed the changed. Just wanted to reach out and say thank you @hrsh7th for all your hard work.

McAuleyPenney:
I think that I may have found some other items of interest for the custom menu. On some occasions, it looks like the left half of the menu is cut off. On other occasions, seemingly when the menu is open during insert mode, you escape to normal, go back to insert, and start typing, the menu returns but with no visible text. When the menu is like this, you can still scroll through it, see documentation, and select an item from the menu. 

https://user-images.githubusercontent.com/59481467/135142164-b6585bd0-2d8c-42fb-82e1-4e5c79f72efa.mp4

sQVe:
I'm curios on how to best handle bindings for the `custom-menu` vs the `native-menu`. I've setup my <kbd>Tab</kbd> key to trigger:

```lua
local tab_complete = function(fallback)
    if cmp.visible() then
        cmp.select_next_item()
    else
        fallback()
    end
end
```

This no longer handles <kbd>Tab</kbd> binding for the native menu. For example when wanting completion for a full line via <kbd>Ctrl+x</kbd><kbd>Ctrl+l</kbd>. How do I best support <kbd>Tab</kbd> and <kbd>Shift+Tab</kbd> on both `native` and `custom` menu?

hrsh7th:
@McAuleyPenney Thank you for debugging.
I can't think of the cause... I didn't meet this problem. How to reproduce? (minimal step is needed)

McAuleyPenney:
> @McAuleyPenney Thank you for debugging. I can't think of the cause... I didn't meet this problem. How to reproduce? (minimal step is needed)

No problem, I'll try and find a minimal step. To be honest, it's happening to me in nearly all documents that I have worked on today. It almost looks like the content of the menu is "sliding" to the left; sometimes the content looks correct, then it is halfway out of the menu, then it is completely gone. Maybe it is another setting or plugin

hrsh7th:
@McAuleyPenney I've prevent events. Could you re-check it?

hrsh7th:
@sQVe I understand the cause. But it's a bit tricky and complicated... Hm...

McAuleyPenney:
@hrsh7th I'm testing with the update you just made right now and it seems like that fixed it 👍🏻  

I was testing without plugins and it _looks_ like it worked without issues when I removed my color scheme, so I don't know what that's about

tzachar:
I am getting some strange behaviour after an editing session has been running for a while.
When getting suggestions, after pressing `<C-c>`, instead of exiting the completion menu a random word gets completed.
I am mapping `<C-c>` to `cmp.mapping.abort()`. 
Just pressing `<Esc>` exists appropriately.

ratheesh:
I am getting following error while using lsp source(using clangd)
`Error executing vim.schedule lua callback: ...im/plugged/nvim-cmp/lua/cmp/view/custom_entries_view.lua:112: invalid option '%-103' to 'format'`
I bisected to the following commit
`22430fb38a90564c9524044064d8e53f601616ae`

hrsh7th:
please update.

ratheesh:
@hrsh7th 
I am using latest HEAD `68c868806f3f33da9a3fad7dd67c53f69abec4ab` on `custom-menu` branch

hrsh7th:
Hm... It's weird. I'm trimming all texts in entry.lua's get_view function.

hrsh7th:
@ratheesh I've updated the formatting pattern. Could you test with this?

ratheesh:
> @ratheesh I've updated the formatting pattern. Could you test with this?

@hrsh7th, With latest commit `c41b1b08848629adf308b2988bc288e09d69eb8a`, it works fine. Thank you.

rhcher:
## Explain the issue
The completion menu does not close as expected with nvim-autopairs plugin but it works well in main branch. So I guess it a issue about custom-menu.

### Steps to reproduce
 1. run `nvim -u vimrc.vim test.lua`
 2. We input like this and we can notice that the completion menu appears.
image

 3. Then we input `)` and the completion menu still exists.
image

### Expected behavior
The completion menu close as in main branch with nvim-autopairs plugin.

### Observed behavior
The completion menu dose not close as expected with nvim-autopairs.

## Minimal working example
test.lua
```lua
x = 1
y = 1

print(x, y)
```

## Minimal vimrc file
vimrc.vim
```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp', { 'branch': 'custom-menu' }
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
Plug 'windwp/nvim-autopairs'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
  },

  sources = {
    { name = "nvim_lsp" },
    { name = "buffer" },
  },
}
EOF

lua << EOF

require'lspconfig'.sumneko_lua.setup {
  capabilities = capabilities,
  cmd = {
    "/home/rhcher/workspace/lua-language-server/bin/Linux/lua-language-server",
    "/home/rhcher/workspace/lua-language-server/main.lua",
  },
}
EOF

lua << EOF
require('nvim-autopairs').setup{}
EOF
```

## Additonal information
nvim version: NVIM v0.6.0-dev+366-g270cc1d70f
I can provide any other information if needed

tzachar:
Hi.

After the source timeout commit, I keep getting the following error:
```
Error executing vim.schedule lua callback: .../site/pack/packer/start/nvim-cmp/lua/cmp/utils/async.lua:47: handle 0x55d132739c20 is already closing    
```

ratheesh:
If `clangd` is used as the LSP server with unimported suggested(denoted by dot and space in the beginning), `Kind Label` and `Source` are misaligned with other items as shown in the picture below. I am using the latest commit. 
clangd_nvim_cmp_bug

hrsh7th:
@ratheesh What is your `ambiwidth` option value? Please paste the output of `echomsg &ambiwidth`

ratheesh:
@hrsh7th , output of `echomsg &ambiwidth` is `single`

hrsh7th:
OK. It's  a bug. I will check it.

gegoune:
Amazing job @hrsh7th! Thank you so so much!

mortezadadgar:
I feel it's faster now!

yujinyuz:
Thanks @hrsh7th I can feel drastic improvements compared to the old one

tami5:
😍 @hrsh7th this is crazy!!! well done you just reach something I don't think anyone can match. Congratulations

oblitum:
[Tried cmp briefly](https://github.com/hrsh7th/nvim-cmp/issues/455#issuecomment-980641982) and it seems the [suggestion for `preselect = 'always'`](https://github.com/hrsh7th/nvim-cmp#what-is-the-equivalence-of-nvim-compes-preselect--always) has no effect after this PR.

PR #228, opened 2021-09-21
Fix: bad function usage in README for UltiSnips expansion

Theory-of-Everything:
This commit fixes an issue with using vsnip#anonymous in the snippet expand function for Ultisnip
The main issue, is that when using `vsnip#anonymous` instead of `UltiSnip#Anon` when calling the `cmp.mapping.confirm` map on a non-snippet option would "crash" cmp, ceasing its normal functionality. Using `UltiSnps#Anon` fixes this issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ lua <<EOF
         -- require('luasnip').lsp_expand(args.body)
 
         -- For `ultisnips` user.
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thx!

PR #240, opened 2021-09-23
Improvement idea: option to disables insert on select item

ericpubu:
Following up on this https://github.com/hrsh7th/nvim-cmp/pull/224#issuecomment-925111914 I started to check the code, and it might make sense to insert on each selection, this might be a design choice, but this would add a configuration option that disables this behavior, making the `select_prev_item()` and  `selec_next_item()`  just move through the menu without inserting until `confirm()` is called.

these are some examples with the config combination

```
  mapping = {
    ["<C-p>"] = cmp.mapping.select_prev_item{behavior = cmp.SelectBehavior.Select},
    ["<C-n>"] = cmp.mapping.select_next_item{behavior = cmp.SelectBehavior.Select},
  },
  experimental = {
    ghost_text = true,
  },

```

https://user-images.githubusercontent.com/18682359/134438815-d2914d1a-ce66-4137-b14d-73bf7ac514ba.mov

```
  mapping = {
    ["<C-p>"] = cmp.mapping.select_prev_item{behavior = cmp.SelectBehavior.Select},
    ["<C-n>"] = cmp.mapping.select_next_item{behavior = cmp.SelectBehavior.Select},
  },
  experimental = {
    ghost_text = false,
  },
```

https://user-images.githubusercontent.com/18682359/134438970-f40bdb6b-8aef-414a-a34f-9a2f739ae31c.mov

Both flags in false, or not changing anything (the default config) would result in the current behavior.

Let me know if this idea makes sense to you,  and if there are some required changes here :)

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -43,6 +43,10 @@ return function()
       end,
     },
 
+    selection = {
+      default_behavior = types.cmp.SelectBehavior.Insert,
+    },
+
     sorting = {
       priority_weight = 2,
       comparators = {
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -69,6 +69,7 @@ return function()
     experimental = {
       custom_menu = true,
       ghost_text = false,
+      disables_insert_on_selection = false
     },
 
     sources = {},
--- a/README.md
+++ b/README.md
@@ -397,6 +397,12 @@ Specify whether to display ghost text.
 
 Default: `false`
 
+#### experimental.disables_insert_on_selection (type: boolean)
+
+Disables insertion on `select_prev_item()` and `select_next_item()`
+if `experimental.ghost_text` is set to `true` it would refresh the ghost text on next/prev item if this is setted to true
+
+Default: `false`
 
 Commands
 ====================
--- a/README.md
+++ b/README.md
@@ -130,8 +130,8 @@ If you want to remove an option, you can set it to `false` instead.
 
 Built in helper `cmd.mappings` are:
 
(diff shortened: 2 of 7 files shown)


hrsh7th:
Thank you.
I think it should be arugment of `cmp.select_next_item` / `select_prev_item`.

ericpubu:
@hrsh7th that makes sense! I made the change

ericpubu:
So giving more thought to this, defaulting to use ghost text when disabling the insert might not be the best idea, so I ended up changing this flag to an enum `behavior` similar to how the `confirmation` option is working now. I updated the PR description to reflect this

lua/cmp/config/default.lua
```diff
@@ -43,6 +43,10 @@ return function()
       end,
     },
 
+    selection = {
+      default_behavior = types.cmp.SelectBehavior.Insert,
+    },
```
hrsh7th, line 48: I think we should remove this option. (`cmp.SelectBehavior` constants is good.)

lua/cmp/config/default.lua
```diff
@@ -69,6 +73,7 @@ return function()
     experimental = {
       custom_menu = true,
       ghost_text = false,
+      disables_insert_on_selection = false
```
hrsh7th, line 76: This is unneeded anymore.

hrsh7th:
I want to merge this PR for the next work. So I will merge it and fix it.

ericpubu:
@hrsh7th cool, thanks!
I just realized I left [this line](https://github.com/hrsh7th/nvim-cmp/blob/custom-menu/lua/cmp/config/default.lua#L78) that needs to be clean when I change it, sorry!
Let me know if I can help with anything else

PR #295, opened 2021-10-08
Force `cursorlineopt=line` for selection highlighting in floating window

hexium310:
This PR fixes #288.

PR #301, opened 2021-10-09
Check whether option exists

hexium310:
This PR fixes an issue where an error occurs when the `cursorlineopt` option doesn't exist (https://github.com/hrsh7th/nvim-cmp/pull/295#issuecomment-939233273).

I forgot that `cursorlineopt` only exists in Neovim nightly. Sorry!


hrsh7th:
LGTM!

PR #310, opened 2021-10-09
adding the pumblend property to floating pum window

tamton-aquib:
setting the blend option for pum menu, by setting default blend to `vim.opt.pumblend:get()`


hrsh7th:
LGTM.

PR #313, opened 2021-10-09
Make the pum window property respect future changes of &pumblend

wookayin:
PR #310 added a support for the pumblend property for the custom
floating completion pop-up menu, but the winblend option is set
globally which doesn't respect future changes (e.g., `set pumblend=..`).

Such window options for the pop-up window that might be changed later
would need to be set when the floating window is being opened.

/cc @tamton-aquib (the author of #310)


Shougo:
When the value is changed?
You should describe why the feature is really needea.

hrsh7th:
LGTM! well organized 👍🏼

wookayin:
> When the value is changed? You should describe why the feature is really needea.

Users can change it manually. For example, you may want to play around with several values to find the best one that would visuallylook the best.

Issue #330, opened 2021-10-11
Documentation not display after floating window update


Xuyuanp:
Before:
![1633944251595](https://user-images.githubusercontent.com/2245664/136766627-19d1d7cb-dc0c-42e1-a911-68cc8149a29c.jpg)

After:
![1633944290397](https://user-images.githubusercontent.com/2245664/136766654-c7b9eb6d-145c-4baa-97e8-d12532647e9d.jpg)

My config:

```lua
local replace_termcodes = vim.api.nvim_replace_termcodes

local M = {}

local function t(key)
    return replace_termcodes(key, true, true, true)
end

function M.setup()
    local cmp = require('cmp')
    local compare = require('cmp.config.compare')

    local WIDE_HEIGHT = 40

    cmp.setup({
        snippet = {
            expand = function(args)
                vim.fn['vsnip#anonymous'](args.body)
            end,
        },
        mapping = {
            ['<C-p>'] = cmp.mapping.select_prev_item(),
            ['<C-n>'] = cmp.mapping.select_next_item(),
            ['<C-Space>'] = cmp.mapping.complete(),
            ['<CR>'] = cmp.mapping.confirm({
                behavior = cmp.ConfirmBehavior.Insert,
                select = false,
            }),
            ['<Tab>'] = function(fallback)
                if cmp.visible() then
                    cmp.select_next_item()
                elseif vim.fn['vsnip#available']() == 1 then
                    vim.fn.feedkeys(t('<Plug>(vsnip-expand-or-jump)'), '')
                else
                    fallback()
                end
            end,
            ['<S-Tab>'] = function(fallback)
                if cmp.visible() then
                    cmp.select_next_item()
                elseif vim.fn['vsnip#available']() == 1 then
                    vim.fn.feedkeys(t('<Plug>(vsnip-jump-prev)'), '')
                else
                    fallback()
                end
            end,
        },
        preselect = cmp.PreselectMode.None,
        sources = {
            { name = 'buffer' },
            { name = 'nvim_lsp' },
            { name = 'nvim_lua' },
            { name = 'vsnip' },
            { name = 'path' },
            { name = 'tmux' },
            { name = 'calc' },
            { name = 'crates' },
        },
        sorting = {
            comparators = {
                compare.kind,
                compare.offset,
                compare.exact,
                compare.score,
                compare.sort_text,
                compare.length,
                compare.order,
            },
        },
    })
end

return M
```

Xuyuanp:
[...]

PR #335, opened 2021-10-12
Added option to set zindex of documentation window

Xuyuanp:
#330


Xuyuanp:
I think it's useful to make zindex configurable when we have other plugin with floating window while typing(like signature help). Sometimes these windows will cover each other.

PR #341, opened 2021-10-12
feat: Vim Docs

tjdevries:
I didn't actually do too much to make this work well, but just showing how it could be possible to generate builtin vim docs from the lua code.

There are some things that we don't yet support in https://github.com/tjdevries/tree-sitter-lua, but could be interesting to explore later.

Anyway, just thought I'd shoot this your way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,3 +22,5 @@ integration:
 	luacheck lua
 	vusted lua
 
+docgen:
+	nvim --headless --noplugin -u scripts/minimal_init.vim -c "luafile ./scripts/docgen.lua" -c 'qa'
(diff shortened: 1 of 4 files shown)


tjdevries:
We are working on a docker image to make it easy to run. I will let you know when we finish.

danymat:
As this PR seems stale, I cannot recommend enough `mini.doc` module: https://github.com/echasnovski/mini.nvim#minidoc

I use it to automatically generate vim help file from Lua comments in Neogen plugin.

I wrote a basic tutorial on how to set up for a plugin: 
- https://github.com/wbthomason/packer.nvim/issues/797#issuecomment-1036084926

PR #362, opened 2021-10-16
RFC: cmdline completion

hrsh7th:
this PR contains the cmdline completion (like wilder.nvim)

It may be overkill because wilder.nvim already exists and it is more useful than this.

```lua
  cmp.setup.cmdline('/', {
    sources = {
      { name = 'buffer' }
    }
  })
  cmp.setup.cmdline(':', {
    sources = {
      { name = 'cmdline' }
    }
  })
```

https://user-images.githubusercontent.com/629908/137590370-302d9839-9259-4a8d-bd05-0ee4a56fd645.mp4


hrsh7th:
I will split this PR as `dot & macro support` and `cmdline completion`.

Issue #274, opened 2021-10-05
Bug: nvim-cmp doesn't auto select when chaining methods


AgentCosmic:
For example given this line:
```
object.get().some_property
                          ^ cursor trying to auto complete here
```
When I press tab to select the entry I want, then continue typing, nvim-cmp deletes everything that I've typed such that it's now just:
```
object.get().
```

But it would work just fine when auto completing the first `get()`. In order to solve this I have to use `ctrl-space`.

I'm using nvim 0.5.1 Window. So far I've only tested it with tsserver.

hrsh7th:
Please minimal repro steps.

AgentCosmic:
Pasting my `init.vim` below. I also noticed it happening for any property/method now, doesn't need to be chained. BTW everything was working fine with nvim-compe.

```
let $ROOT = expand('D:/Applications/Neovim/nvim')
set runtimepath+=$ROOT
let $HOME = $ROOT . '/home'

runtime mswin.vim

set termguicolors

call plug#begin('$HOME/plugged')
Plug 'neovim/nvim-lspconfig'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-path'
Plug 'hrsh7th/nvim-cmp'
call plug#end()


" LSP
lua << EOF
local nvim_lsp = require('lspconfig')

-- Use an on_attach function to only map the following keys after the language server attaches to the current buffer
local on_attach = function(client, bufnr)
	local function buf_set_keymap(...) vim.api.nvim_buf_set_keymap(bufnr, ...) end
	local function buf_set_option(...) vim.api.nvim_buf_set_option(bufnr, ...) end
	--Enable completion triggered by <c-x><c-o>
	buf_set_option('omnifunc', 'v:lua.vim.lsp.omnifunc')
end

-- this is where we install all the language servers
local lsp_bins = vim.api.nvim_eval("$ROOT . '\\lsp'")

-- npm i  typescript-language-server
nvim_lsp.tsserver.setup{
	on_attach = function(client, bufnr)
		client.resolved_capabilities.document_formatting = false	
		on_attach(client, bufnr)
	end,
	cmd = { lsp_bins .. '\\node_modules\\.bin\\typescript-language-server.cmd', '--stdio' },
	capabilities = require('cmp_nvim_lsp').update_capabilities(vim.lsp.protocol.make_client_capabilities())
}
EOF


[...]

PR #439, opened 2021-10-30
README: clarify that a snippet plugin is required

bradleymackey:
I don't use a snippet plugin normally and I was running to issues with autocompletion until I came across [this comment](https://github.com/hrsh7th/nvim-cmp/issues/274#issuecomment-939500416).

Therefore, I think it's best to clearly document that a snippet engine is required at the moment. Adds comment to suggested sample and the FAQ.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ lua <<EOF
 
   cmp.setup({
     snippet = {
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM.

PR #447, opened 2021-10-31
Fix the native menu not respecting cmp.PreselectMode.None when the first completion item has preselect = true

dmitmel:


hrsh7th:
LGTM. Thank you!

Issue #445, opened 2021-10-31
Question regarding changes to text before `callback` is called


PatOConnor43:
**Describe the bug**
Hi :wave: I was trying to build a plugin that would autocomplete packages within a programming language's dependency spec file (Dart's pubspec.yaml file to be specific). As the user types I would _like_ to be able to queue a request to the search API and then populate items back to `cmp`. I wanted to do this asynchronously and also be able only query the API after the user has paused typing. The problem is that it seems like `cpm` expects the custom source to always call the `callback` before new calls to the defined `complete` function gets called again.

**Minimal config based on [this](https://github.com/hrsh7th/nvim-cmp/blob/main/utils/vimrc.vim)**

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"

local source = {}
source.new = function()
    local self = setmetatable({}, { __index = source })
    return self
end

source.is_available = function()
    return true
end

function source.get_debug_name()
    return 'test'
end

function source.resolve(self, completion_item, callback)
    callback(completion_item)
end

function source.execute(self, completion_item, callback)
    callback(completion_item)
end

function source.get_trigger_characters(self, config)
    return {'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k', 'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', '_'}
end

function source.complete(self, params, callback)
    
    local cur_line = params.context.cursor_line
    local items = {}

    -- Look for the first word in the line
[...]

PR #450, opened 2021-10-31
Added FETCHING to the get_sources call

PatOConnor43:
This is just a draft PR to show the changes I've made. I don't expect this to get merged. 

Related to #445


hrsh7th:
I understand you are understanding nvim-cmp's mechanism.

However, I have other solution for it.

I think we should handle `isIncomplete` with empty items like the following.

```
callback({ isIncomplete = true, items = {} })
```

Currently, nvim-cmp does not accept isIncomplete=true in this case. So I will fix it.

PR #460, opened 2021-11-02
Add documents

hrsh7th:


doc/cmp.txt
```diff
@@ -0,0 +1,265 @@
...
+
+
+
+==============================================================================
+Abstract                                                          *cmp-abstract*
+
+This is nvim-cmp's document.
+
+1. The `cmp` object is the return value from `require('cmp')`.
+2. Some type-definition can be found via `:CmpTypes`
```
hrsh7th, line 27: It should be removed.

hrsh7th:
I want to auto alignment....

hrsh7th:
It is wip, but I think it can be improved gradually, so I will merge it.

PR #471, opened 2021-11-03
fix(ghost_text): generate it using chars instead of bytes

muniter:
Previously if the completion entry text had a multibyte character the
ghostly text renderd improperly, for example:

The cursor position is "|"

### 1. Example
In this case the ghost text was spliting a multibyte character.

screenshot_20211103_164346
screenshot_20211103_164803

### 2. Example
In this case the multibyte had passed messing up the indexing therefore presentin the last character of the entry as ghost text.

screenshot_20211103_164332
screenshot_20211103_164815

### DIFF
--- a/lua/cmp/view/ghost_text_view.lua
+++ b/lua/cmp/view/ghost_text_view.lua
@@ -26,19 +26,15 @@ ghost_text_view.new = function()
         return
       end
 
-      local cursor = vim.api.nvim_win_get_cursor(0)
-      if string.sub(vim.api.nvim_get_current_line(), cursor[2] + 1) ~= '' then
+      local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+      local line = vim.api.nvim_get_current_line()
+      if string.sub(line, col + 1) ~= '' then
         return
       end
 
-      local diff = 1 + cursor[2] - self.entry:get_offset()
-      local text = self.entry:get_insert_text()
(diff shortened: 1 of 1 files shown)


lua/cmp/view/ghost_text_view.lua
```diff
@@ -51,6 +47,20 @@ ghost_text_view.new = function()
...
+ghost_text_view.text_gen = function(self, line, cursor_col)
+  local text
+  local cword = string.sub(line, self.entry:get_offset(), cursor_col)
+  local cword_char_len = vim.fn.strchars(cword)
+  local entry_text = self.entry:get_insert_text()
+  if self.entry.completion_item.insertTextFormat == types.lsp.InsertTextFormat.Snippet then
+    entry_text = vim.lsp.util.parse_snippet(entry_text)
+  end
+  -- Missing characters to complete the entry text
+  text = vim.fn.strcharpart(str.oneline(entry_text), cword_char_len)
```
hrsh7th, line 60: Could you use `vim.str_byteidx` instead of `vim.fn.strcharpart`?

muniter: Done

muniter:
@hrsh7th Changed the implementation to only use `vim.str_{utf,byte}index`.

Also heads up on the following just merged functions in neovim that will be very helpful for more features in cmp.

- https://github.com/neovim/neovim/pull/16129
- https://github.com/neovim/neovim/pull/16124

hrsh7th:
LGTM. Thank you~!

PR #474, opened 2021-11-04
Escape <Bar> when setting indentkeys

mhoogendoorn:
In `keymap.autoindent`, `!^F` is added to `indentkeys`, then `<C-f>` is
fed.  In the end `indentkeys` is reset to its original value.

We need to escape any `|` in `indentkeys` when setting the option
because otherwise they are seen as the command separator.


mhoogendoorn:
For example [PhilT/vim-fsharp](https://github.com/PhilT/vim-fsharp) has an indentkeys that contains `|` characters.

hrsh7th:
LGTM!

PR #493, opened 2021-11-08
Link to menu appearance page in wiki

gajzl:
Added instruction for icons and source without extra plugin.


hrsh7th:
Looks very good. Thank you!

Issue #327, opened 2021-10-11
Feature: Item Kind Highlight groups


danielnehrig:
It would be cool if we could specify/overwrite for each Item Kind our own Highlights
So for instance item kind Function is 'Blue' and snippet is 'orange'

hrsh7th:
How to realize it? If you have an idea, feel free to post it.
(CmpKindFunction, CmpKindInterface ...)?

danielnehrig:
Correct kind of like this:
`CmpKind .. kind_name`

on the setup function we could introduce a table:
```
kind = {
  Function = "#3d3d3d",
  Snippet = "...",
  ...
}
```
The table is used to compose highlights

danielnehrig:
thanks!

Dan7h3x:
> It would be cool if we could specify/overwrite for each Item Kind our own Highlights So for instance item kind Function is 'Blue' and snippet is 'orange'

I do it by creating custom highlight groups
https://github.com/Dan7h3x/NvimPy

PR #509, opened 2021-11-11
feat: custom highlight on kind item #327

windwp:
hi
my pr allow user can change highlight on lsp kind.

#327 

```lua
--sample

local lsp_kinds = {
    Text                  = {''  , 'TSText'        } ,
    Method                = {'ƒ'  , 'TSMethod'      } ,
    Function              = {'ƒ'  , 'TSFunction'    } ,
    Constructor           = {''  , 'TSConstructor' } ,
    Field                 = {'⌘'  , 'TSField'       } ,
    Variable              = {''  , 'TSVariable'    } ,
    Class                 = {''  , 'TSConstructor' } ,
    Interface             = {''  , 'TSAnnotation'  } ,
    Module                = {''  , 'TSConstant'    } ,
    Property              = {''  , 'TSProperty'    } ,
    Unit                  = {''  , 'TSTag'         } ,
    Value                 = {''  , 'TSValue'       } ,
    Enum                  = {'了' , 'TSMethod'      } ,
    Keyword               = {''  , 'TSKeyword'     } ,
    Snippet               = {'﬌'  , 'TSLabel'       } ,
    Color                 = {''  , 'TSBoolean'     } ,
    File                  = {''  , 'TSSymbol'      } ,
    Reference             = {''  , 'TSReference'   } ,
    Folder                = {''  , 'Directory'     } ,
    EnumMember            = {''  , 'TSPunctSpecial'} ,
    Constant              = {''  , 'TSConstant'    } ,
    Struct                = {''  , 'TSStruct'      } ,
    Event                 = {''  , 'TSEvent'       } ,
    Operator              = {''  , 'TSOperator'    } ,
    TypeParameter         = {''  , 'TSType'        } ,
}

cmp.setup({
    formatting={
      format = function(_, vim_item)
        vim_item.kind_highlight = lsp_kinds[vim_item.kind][2]
        vim_item.kind = lsp_kinds[vim_item.kind][1]
        return vim_item
      end,
    }
})
```


hrsh7th:
This implementation may also be good but I was thinking of a different way.
I was thinking of defining highlights for all kind in `plugin / cmp.lua`. (This method has the advantage that colors can be defined in the color scheme.)

```
highlight CmpItemKind ...
highlight link CmpItemKindFunction CmpItemKind
highlight link CmpItemKindInterface CmpItemKind
...
```

_google translated_

windwp:
i already think about it but it need define many highlight group and it need a colorscheme plugin to update.

i already use it and mapping an lsp item kind to a treesitter highlight group is easy to recognize.

hrsh7th:
I think your solution will not allow you to specify highlight groups in your color scheme.

My solution allows it, but we can't programmatically specify highlight groups from the plugin.

I think my solution is more vim way.

DimitrisMilonopoulos:
I would say I agree with @hrsh7th. I was thinking about this today, so I just went over the PRs as I was sure that somebody else would have thought of it. 

Yes if these Highlight groups are not defined then we can fallback to the original Highlight for the Kinds.

PR #514, opened 2021-11-12
Fire change events when selecting different items with native menu completion

Lyude:
Previously we didn't do this, which resulted in the documentation window
not being shown when switching through different completion items. Fix
this by firing the change event in select_next_item and select_prev_item.

### DIFF
--- a/lua/cmp/view/native_entries_view.lua
+++ b/lua/cmp/view/native_entries_view.lua
@@ -118,21 +118,27 @@ native_entries_view.preselect = function(self, index)
 end
 
 native_entries_view.select_next_item = function(self, option)
+  local callback = function()
+    self.event:emit('change')
+  end
   if self:visible() then
     if (option.behavior or types.cmp.SelectBehavior.Insert) == types.cmp.SelectBehavior.Insert then
-      feedkeys.call(keymap.t('<C-n>'), 'n')
+      feedkeys.call(keymap.t('<C-n>'), 'n', callback)
     else
-      feedkeys.call(keymap.t('<Down>'), 'n')
+      feedkeys.call(keymap.t('<Down>'), 'n', callback)
     end
   end
 end
 
 native_entries_view.select_prev_item = function(self, option)
+  local callback = function()
+    self.event:emit('change')
+  end
   if self:visible() then
     if (option.behavior or types.cmp.SelectBehavior.Insert) == types.cmp.SelectBehavior.Insert then
-      feedkeys.call(keymap.t('<C-p>'), 'n')
+      feedkeys.call(keymap.t('<C-p>'), 'n', callback)
     else
-      feedkeys.call(keymap.t('<Up>'), 'n')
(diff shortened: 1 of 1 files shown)


hrsh7th:
I don't think it's correct. See https://github.com/hrsh7th/nvim-cmp/pull/514/files#diff-08b3a05fbbdb22b51fc7910a2977836da2796d89b00fe66828cedf0e31d17ccfR24

The native menu will fires the `CompleteChanged` event.

Lyude:
> I don't think it's correct. See https://github.com/hrsh7th/nvim-cmp/pull/514/files#diff-08b3a05fbbdb22b51fc7910a2977836da2796d89b00fe66828cedf0e31d17ccfR24
> 
> The native menu will fires the `CompleteChanged` event.

Sorry, I see you linked back to the commit diff for this MR but I'm not sure I'm totally following. Are you saying that nvim should be firing the `CompleteChanged` event every time the native completion popup's selection is changed, which would presumably handle firing off the `changed` event here:

```lua
  autocmd.subscribe('CompleteChanged', function()
    self.event:emit('change')
  end)
```

However, this doesn't appear to be the case with nvim 0.5.1, where I only ever see the `CompleteChanged` event get emitted when the completion initially begins, or when the text which triggered the completion gets changed while the popup is still open. Could this possibly be some sort of neovim bug then?

hrsh7th:
Hm... I think if we use `cmp.SelectBehavior.Select`, the vim does not fires the event.

OK. I'll review it.

lua/cmp/view/native_entries_view.lua
```diff
@@ -133,6 +134,7 @@ native_entries_view.select_prev_item = function(self, option)
     else
       feedkeys.call(keymap.t('<Up>'), 'n')
     end
+    self.event:emit('change')
   end
```
hrsh7th, line 138:
```
    local callback = function()
      self.event:emit('change')
    end
    if (option.behavior or types.cmp.SelectBehavior.Insert) == types.cmp.SelectBehavior.Insert then
      feedkeys.call(keymap.t('<C-p>'), 'n', callback)
    else
      feedkeys.call(keymap.t('<Up>'), 'n', callback)
    end
```

lua/cmp/view/native_entries_view.lua
```diff
@@ -123,6 +123,7 @@ native_entries_view.select_next_item = function(self, option)
     else
```
hrsh7th, line 123: We should remove `CompleteChanged` in the above of lines (maybe)

Lyude: Are you talking about:

```rust
  autocmd.subscribe('CompleteChanged', function()
    self.event:emit('change')
  end)
```

I think we would still need this for when the actual content of the popup menu has changed, no?

hrsh7th:
Sorry for the lately reply.
I can merge this.

PR #525, opened 2021-11-15
WIP: Improve `entry:get_offset` calculation

hrsh7th:


hrsh7th:
I have to organize entry.lua codes before merge this

PR #527, opened 2021-11-16
Automagically enable LSP completion in vimrc

jamietanna:
To make it easier for folks to add new servers, without needing to add
the configuration in their vimrc, too, we can instead use the
`lspconfig`'s `available_servers` method to list all servers, and
automagically apply our capabilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,10 +122,13 @@ lua <<EOF
 
   -- Setup lspconfig.
   local capabilities = require('cmp_nvim_lsp').update_capabilities(vim.lsp.protocol.make_client_capabilities())
-  -- Replace <YOUR_LSP_SERVER> with each lsp server you've enabled.
-  require('lspconfig')['<YOUR_LSP_SERVER>'].setup {
-    capabilities = capabilities
-  }
+  -- Automagically enable each lsp server that's available
+  local lspconfig = require('lspconfig')
+  for key,server in pairs(lspconfig.available_servers()) do
+    lspconfig[server].setup {
+      capabilities = capabilities
+    }
+  end
 EOF
 ```
 


mjlbach:
Please do not encourage people to do this.

jamietanna:
> Please do not encourage people to do this.

Sorry why not? Still new to Neovim, so not sure if there's anything security / performance wise that'd be a problem?

mjlbach:
> Sorry why not? 

Did you try it?

FWIW I am the maintainer of lspconfig, while you could shoe-horn a solution that uses available_servers into working by first populating the configs table, I'd rather people reference the extensive readme, wiki, and documentation (:help lspconfig) in lspconfig.

windwp:
you call a setup of all server but people only use a few server. it increase a startup time.
if you open a file has a server setup but you don't install lsp command it will display a warning.

Maybe you can make a PR on lspconfig to create a hook for modify capabilities from plugin. nvim-cmp only call that hook and we can remove that step on nvim-cmp Readme.

Shougo:
I think users should choose/config LSP servers instead of the auto.

If the feature is really needed, it should be added FAQ or other documentation(wiki?) instead of README.

dmitmel:
> you call a setup of all server but people only use a few server. it increase a startup time.

Doesn't `available_servers()` return a list of configured servers? <https://github.com/neovim/nvim-lspconfig/blob/a209627886c3edc6926bede732543f9c09aaf22c/plugin/lspconfig.vim#L8>

hrsh7th:
I think the `available_servers` is useless to this situation.
because the setup function can accept the specific configuration per each servers but this abstraction disallow to configure that.

PR #529, opened 2021-11-16
Add reason to complete function

### LINKED ISSUE #528: `keyword_length` when reason is manual
Currently, the code does not check `keyword_length` when completion is triggered manually.
https://github.com/hrsh7th/nvim-cmp/blob/4d800fbcab86e545b65306b0c4c0fde03f2a62b8/lua/cmp/source.lua#L249-L252

I think it should still check.
When I have different `keyword_length` per source, even when I trigger completion manually, I still expect only sources to show that have a lower `keyword_length` than the current word length.

I can make a PR to change it, but I wanted to check first if you agree

lukas-reineke:
Allow to specify the reason (or any other context options) when calling `cmp.mapping.complete`

fix #528


hrsh7th:
Perfect. Thank you!

PR #537, opened 2021-11-17
Add documentation.delay option to config

clktmr:
This option can be used to reduce induced load on slow language servers
or just to delay the documentation popup.

### DIFF
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -180,11 +180,15 @@ view._get_entries_view = function(self)
 
   if c.experimental.native_menu then
     self.native_entries_view.event:on('change', function()
+      self.on_entry_change.timeout = c.documentation.delay
+      self.on_entry_change.stop()
       self:on_entry_change()
     end)
     return self.native_entries_view
   else
     self.custom_entries_view.event:on('change', function()
+      self.on_entry_change.timeout = c.documentation.delay
+      self.on_entry_change.stop()
       self:on_entry_change()
     end)
     return self.custom_entries_view
@@ -221,7 +225,7 @@ view.on_entry_change = async.throttle(
       self.ghost_text_view:hide()
     end
   end),
-  20
+  config.get().documentation.delay
 )
 
 return view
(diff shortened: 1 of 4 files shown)


lua/cmp/view.lua
```diff
@@ -180,11 +180,15 @@ view._get_entries_view = function(self)
...
     self.native_entries_view.event:on('change', function()
+      self.on_entry_change.timeout = c.documentation.delay
+      self.on_entry_change.stop()
       self:on_entry_change()
     end)
     return self.native_entries_view
   else
     self.custom_entries_view.event:on('change', function()
+      self.on_entry_change.timeout = c.documentation.delay
+      self.on_entry_change.stop()
```
hrsh7th, line 191: If we call the `cmp.AsyncThrottle.stop` this isn't work as `throttle` but `debounce`.
Why you are calling the `stop` here?

clktmr: I have a *very* slow language server which needs several seconds to respond with the documentation. I had the best experience if the server was only requested once I stopped on a completion item. But just increasing throttling is also an improvement, so I can remove the call to stop if you think it's unreasonable.

clktmr: Put differently, throttle forces me to use a delay higher than the average response time of the server. Otherwise requests stack up and it becomes unresponsive for some time. Debounce allows me to use a delay slightly above my typing speed, which in my case is much lower.

hrsh7th: At least, this must be a throttle due to preselect feature at the moment... Hm...

clktmr: Can you clarify? I don't see how this breaks the preselect feature. The documentation window should still open after `delay` if `on_entry_change` was triggered by an preselect.

hrsh7th: The documentation window will be as you type updated if the `preselect` feature is enabled.

The debouncing is deterioration this behavior.

clktmr: Ah, I see. Since this change is mainly driven by this specific, extraordinarily slow language server, I will stick with my personal fork for now.

PR #538, opened 2021-11-17
Fix document about scrolling docs backwards

hexium310:
A better key mapping, corresponding to `<C-f>`, for scrolling docs backwards is `<C-b>` because it is used for scrolling backwards in Vim (`:h CTRL-B`).

If you have purposely written `<C-d>`, close this PR. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,7 +83,7 @@ lua <<EOF
       end,
     },
     mapping = {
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM.

PR #550, opened 2021-11-18
Fix possible nil argument to string.byte

smjonas:
This fixes the following error that occurred when trying to expand a `luasnip` choicenode snippet at the start of a line:
```lua
Error executing vim.schedule lua callback:
.../nvim/site/pack/packer/start/nvim-cmp/lua/cmp/source.lua:240:
bad argument #1 to 'byte' (string expected, got nil)
```


hrsh7th:
Thank you. LGTM

PR #584, opened 2021-11-27
Implement kind highlights

hrsh7th:
Fixes https://github.com/hrsh7th/nvim-cmp/issues/327

Related https://github.com/hrsh7th/nvim-cmp/pull/509

@windwp What do you think about it?


windwp:
It create 48 highlight group for a small feature even if user don't use it.
but it is a normal way to create highlight so i think it is better.

hrsh7th:
Thank you :bow:

wookayin:
Is this a breaking? `CmpItemKind` stopped working. No proper doc was made in the commit, in this issue, in the README. 

My config:
```
hi CmpItemKind     guifg=something
```

It doesn't apply and only the default highlights are shown. Shouldn't it work as the default highlight for ALL kinds?

PR #591, opened 2021-11-29
fix: default highlights overriding existing highlights

Iron-E:
`highlight! default` overrides the existing highlight in order to set a default link. So any highlight group that exists before will be wiped over.

To fix this I have changed `highlight! default` to `highlight default`, which only sets the highlight group if it wasn't already defined.


hrsh7th:
I don't understand but it looks good to me.

hrsh7th:
Thank you!

PR #607, opened 2021-12-04
Add ghost_visible()

jcorbin:
This enables things like a super-tab that expands ghost text when possible:
```lua
    ['<Tab>'] = cmp.mapping(function(fallback)
      if cmp.ghost_visible() then
        cmp.confirm({ select = true })
      elseif cmp.visible() then
        cmp.select_next_item()
      elseif snip.expand_or_jumpable() then
        snip.expand_or_jump()
        cmp.complete()
      else
        fallback()
      end
    end, {"i", "s"}),
```

### DIFF
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -65,6 +65,11 @@ cmp.visible = function()
   return cmp.core.view:visible() or vim.fn.pumvisible() == 1
 end
 
+---Return view ghost text is visible or not.
+cmp.ghost_visible = function()
+  return cmp.core.view:ghost_visible()
+end
+
 ---Get current selected entry or nil
 cmp.get_selected_entry = function()
   return cmp.core.view:get_selected_entry()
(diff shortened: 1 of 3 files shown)


hrsh7th:
I don't want to add this...

Hm... I guess what you want is `if the cursor not in the middle of current line, should expand first entry by pressing Tab`, right?

jcorbin:
Hmm, that's an interesting observation
- supposing most of the time that I want this it is to confirm/start a snippet expansion, as a dual to the fact that <Tab> is also my snippet next/jump mapping
- and most of that time is going to begin on a new, maybe indented, line

That said, I think I've since realized that ghost text almost always shows up once a reasonably long first match presents...

I suppose what I'm really looking for here would be a scheme more like:
- ghost text only
- display menu with a keybind if desired
- but mostly use typing more to clarify the ghost match, sans menu until invoked

So to get that, I think what I'd like to try is:
- auto ghost text completion
- Tab should be able to "show menu if matches exist, but it isn't shown yet" and then "advance menu if shown"

So yeah, this specific function isn't what I'm looking for after all, more research required ;-)

PR #622, opened 2021-12-07
Readme: improve readability of lua code block in setup section

smjonas:
I just noticed this while reading the setup section.

Compare

 </img>

with

 </img>

Especially the comments look nicer imo.


hrsh7th:
OK. I think it is not correct annotation but it's convinient.

PR #639, opened 2021-12-12
feat: make throttle time configurable

HansPinckaers:
For me nvim-cmp feels a lot more snappy when lowering the throttle_time in core.lua. I do not notice an increase in flickering or something else on my device. As such, I think it makes sense to make this a configurable option.

Thanks for the work on this amazing plugin!

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -120,6 +120,7 @@ return function()
     experimental = {
       native_menu = false,
       ghost_text = false,
+      throttle_time = 120,
     },
 
     sources = {},
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -14,7 +14,6 @@ local api = require('cmp.utils.api')
 local event = require('cmp.utils.event')
 
 local SOURCE_TIMEOUT = 500
-local THROTTLE_TIME = 120
 
 ---@class cmp.Core
 ---@field public suspending boolean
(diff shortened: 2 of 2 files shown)


smjonas:
In case this gets merged: Which value for throttle time would you recommend? :)

HansPinckaers:
Currently, I have it set to 5 😉

hrsh7th:
Why did you choose to use the experimental key?
I can't merge them unless a good API is proposed.
I don't have any idea where to place it either.

deepl translated

HansPinckaers:
Maybe we can put it under `completion`?

MurdeRM3L0DY:
Hello. How about putting it in:
```lua
require('cmp').setup {
  ...
  core = {
    throttle = 25,
    debounce = 50,
  },
  ...
}
```

MurdeRM3L0DY:
Hello @hrsh7th. Would you merge if I opened another pr for this?

hrsh7th:
I can merge if the better configuration group nane was suggested.

MurdeRM3L0DY:
@hrsh7th What do you think about putting it under `completion` config?

hrsh7th:
Currently, `DEBOUNCE_TIME` is used for debouncing and displaying the `completion response`.

For example, the `calc` source and the` path` source are synchronous sources.
If the user calls the `calc` source and the` path` source at the same time, the filter and display process will be duplicated.
(So I implemented the `DEBOUNCE_TIME` value).

I think we shouldn't call it as `completion`.

I think it should be called as `latency` or `performance` or etcetc...

hrsh7th:
The other contributor solved the my concerned point on other PR. Thank you for contribution.

PR #641, opened 2021-12-12
Fix rendering of completion items with wide characters in the custom entries window

dmitmel:
What I mean by "wide characters" are characters which require more than one column to be rendered. Examples of such characters include:

- CJK characters
- Emoji
- The `<Tab>` character
- Special unprintable characters (for example, from 0 to 31 in the ASCII table) - this one is Vim-specific because Vim renders them as `<xx>` or `^X` (see [`:help 'isprint'`](https://neovim.io/doc/user/options.html#'isprint'))

Because these characters are wide, using `vim.str_utfindex` is not an accurate way of getting the width of a string as it will be rendered on the screen. The solution is to use the Vimscript function `strwidthdisplay`, which is what this PR is for. However, most of the changes in this PR were done for calling `strwidthdisplay` within the entries window's buffer.

Here are some screenshots to better demonstrate the bug:

Before:

image

After:

image

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -198,27 +198,34 @@ entry.is_deprecated = function(self)
 end
 
 ---Return view information.
+---@param suggest_offset number
+---@param entries_buf number The buffer this entry will be rendered into.
 ---@return { abbr: { text: string, bytes: number, width: number, hl_group: string }, kind: { text: string, bytes: number, width: number, hl_group: string }, menu: { text: string, bytes: number, width: number, hl_group: string } }
-entry.get_view = function(self, suggest_offset)
+entry.get_view = function(self, suggest_offset, entries_buf)
   local item = self:get_vim_item(suggest_offset)
-  return self.cache:ensure({ 'get_view', self.resolved_completion_item and 1 or 0 }, function()
+  return self.cache:ensure({ 'get_view', self.resolved_completion_item and 1 or 0, entries_buf }, function()
     local view = {}
-    view.abbr = {}
-    view.abbr.text = item.abbr or ''
-    view.abbr.bytes = #view.abbr.text
-    view.abbr.width = vim.str_utfindex(view.abbr.text)
-    view.abbr.hl_group = self:is_deprecated() and 'CmpItemAbbrDeprecated' or 'CmpItemAbbr'
-    view.kind = {}
-    view.kind.text = item.kind or ''
-    view.kind.bytes = #view.kind.text
-    view.kind.width = vim.str_utfindex(view.kind.text)
-    view.kind.hl_group = 'CmpItemKind' .. types.lsp.CompletionItemKind[self:get_kind()]
-    view.menu = {}
-    view.menu.text = item.menu or ''
-    view.menu.bytes = #view.menu.text
-    view.menu.width = vim.str_utfindex(view.menu.text)
-    view.menu.hl_group = 'CmpItemMenu'
-    view.dup = item.dup
+    -- The result of vim.fn.strdisplaywidth depends on which buffer it was
+    -- called in because it reads the values of the option 'tabstop' when
+    -- rendering <Tab> characters.
+    vim.api.nvim_buf_call(entries_buf, function()
+      view.abbr = {}
+      view.abbr.text = item.abbr or ''
+      view.abbr.bytes = #view.abbr.text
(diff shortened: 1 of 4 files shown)


hrsh7th:
It's almost good to me.

I'm wonder the overhead of `nvim_buf_call`. (If completion item is over 1k, the nvim_buf_call will be called over 1k too)

hrsh7th:
But I can merge this. Thank you~!

dmitmel:
> I'm wonder the overhead of `nvim_buf_call`. (If completion item is over 1k, the nvim_buf_call will be called over 1k too)

You are right, the overhead of `nvim_buf_call` is significant in this case. I've done a quick benchmark with this:

```lua
diff --git a/lua/cmp/entry.lua b/lua/cmp/entry.lua
index a4590cd..c983433 100644
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -205,10 +205,13 @@ entry.get_view = function(self, suggest_offset, entries_buf)
   local item = self:get_vim_item(suggest_offset)
   return self.cache:ensure({ 'get_view', self.resolved_completion_item and 1 or 0, entries_buf }, function()
     local view = {}
+    local outer_time_start = vim.loop.hrtime()
+    local inner_time_elapsed
     -- The result of vim.fn.strdisplaywidth depends on which buffer it was
     -- called in because it reads the values of the option 'tabstop' when
     -- rendering <Tab> characters.
     vim.api.nvim_buf_call(entries_buf, function()
+      local inner_time_start = vim.loop.hrtime()
       view.abbr = {}
       view.abbr.text = item.abbr or ''
       view.abbr.bytes = #view.abbr.text
@@ -225,7 +228,15 @@ entry.get_view = function(self, suggest_offset, entries_buf)
       view.menu.width = vim.fn.strdisplaywidth(view.menu.text)
       view.menu.hl_group = 'CmpItemMenu'
       view.dup = item.dup
+      inner_time_elapsed = vim.loop.hrtime() - inner_time_start
     end)
+    local outer_time_elapsed = vim.loop.hrtime() - outer_time_start
+    local overhead = outer_time_elapsed - inner_time_elapsed
+    printf(
+      'outer:%.3fms inner:%.3fms overhead:%.3fms percent:%.2f%%',
+      outer_time_elapsed / 1e6, inner_time_elapsed / 1e6, overhead / 1e6,
+      (overhead / outer_time_elapsed) * 100
+    )
     return view
   end)
 end
```

And it shows that `nvim_buf_call` takes around 70-80% of execution time of the whole `get_view` function. I'll make a follow-up PR.

dmitmel:
On a second thought... It adds 70-80% to what is less than a millisecond, and if I measure the impact with

```diff
diff --git a/lua/cmp/view/custom_entries_view.lua b/lua/cmp/view/custom_entries_view.lua
index 92cbe37..ca92a5e 100644
--- a/lua/cmp/view/custom_entries_view.lua
+++ b/lua/cmp/view/custom_entries_view.lua
@@ -117,8 +117,11 @@ custom_entries_view.open = function(self, offset, entries)
   local dedup = {}
   local preselect = 0
   local i = 1
+  local total_time = 0
   for _, e in ipairs(entries) do
+    local time_start = vim.loop.hrtime()
     local view = e:get_view(offset, entries_buf)
+    total_time = total_time + (vim.loop.hrtime() - time_start)
     if view.dup == 1 or not dedup[e.completion_item.label] then
       dedup[e.completion_item.label] = true
       self.column_width.abbr = math.max(self.column_width.abbr, view.abbr.width)
@@ -132,6 +135,7 @@ custom_entries_view.open = function(self, offset, entries)
       i = i + 1
     end
   end
+  print(total_time / 1e6)
   vim.api.nvim_buf_set_lines(entries_buf, 0, -1, false, lines)
 
   local width = 0
```

By typing `/usr/bin/` (my `/usr/bin` directory contains ~5k files, but the `get_view` is invoked only 200 times for the first 200 entries), then the overhead of using `nvim_buf_call` on average is 1-2 ms at most. Putting `nvim_buf_call` everywhere will make the code messy, so I think that a very slight overhead is acceptable in exchange for clean code.

hrsh7th:
Thank you for your investigation.
If we optimize this code, entry: get_view will have an implicit prerequisite.
I don't think there is a problem in the current situation, so it's okay as it is.

_google translated_

PR #669, opened 2021-12-26
Change cmp#ready to CmpReady and document autocmds better

tzx:
I was bored, so I browsed the codebase to learn more about how nvim-cmp works.

I changed `cmp#ready` to `CmpReady` because I was confused with the naming convention; I thought it was an autoloaded function. I think the naming convention for User event autocommand patterns is upper camel-case. I also added a description in the docs on how to use these autocommands.

Please feel free to disregard if you disagree with any of these changes.


hrsh7th:
LGTM

Issue #667, opened 2021-12-25
Disable completion when in a comment


mbpowers:
I don't know if it's just me, but this would be a useful feature IMO.

Shougo:
It is useful feature, but comment parsing is very difficult.

mbpowers:
Since the syntax highlighting knows when you are in a comment, is there any way that this could  piggyback off that?

Shougo:
The default syntax highlight knows it, but if treesitter is enabled it is hard.

dmitmel:
```lua
cmp.setup({
  enabled = function()
    local lnum, col = vim.fn.line('.'), math.min(vim.fn.col('.'), #vim.fn.getline('.'))
    for _, syn_id in ipairs(vim.fn.synstack(lnum, col)) do
      syn_id = vim.fn.synIDtrans(syn_id) -- Resolve :highlight links
      if vim.fn.synIDattr(syn_id, 'name') == 'Comment' then
        return false
      end
    end
    return true
  end,
```

Shougo:
Yes, it only works for non treesitter syntax

max397574:
I'm working on a function for this
I'll open a pr soon

PR #676, opened 2021-12-28
add function to check if in comment

max397574:
closes https://github.com/hrsh7th/nvim-cmp/issues/667

This allows something like
```lua
enabled = function()
  return not cmp.inside_comment()
end
```


dmitmel:
I think this should go into the `cmp.config` namespace, like `cmp.config.sources`.

hrsh7th:
@max397574 Thank you for your contribution.
@dmitmel is correct. we should move this to `cmp.config.*` namespace (it means `configuration utilities`).

And I think we should check treesitter existence.

I prefer `cmp.config.context.in_comment_syntax` and `cmp.config.context.in_comment_treesitter` like name.
(Please other suggestions!)

dmitmel:
I suggest an API like this:

1. `cmp.config.context.in_treesitter_capture('comment')` - takes the name of the capture instead of the name of a highlight group to avoid calling a private function to do conversion (no idea if I am using the term "capture" correctly, I don't use treesitter)
2. `cmp.config.context.in_syntax_group('Comment')` - implemented with the function I posted in the bug report.

I don't want to restrict this to just finding comments.

hrsh7th:
Very good suggestion.

max397574:
Great idea imo
And it would also be easy to implement.
What do you think @hrsh7th ?

hrsh7th:
I can accept the @dmitmel suggestion 👍🏼

max397574:
you can now do this:
```lua
  enabled = function()
    if require"cmp.config.context".in_treesitter_capture("comment")==true or require"cmp.config.context".in_syntax_group("Comment") then
      return false
    else
      return true
    end
end
```

hrsh7th:
LGTM. Thank you!

lua/cmp/config/context.lua
```diff
@@ -0,0 +1,69 @@
...
+  local buf = vim.api.nvim_get_current_buf()
+  local row, col = unpack(vim.api.nvim_win_get_cursor(0))
+  row = row - 1
+  if vim.api.nvim_get_mode().mode == 'i' then
+    col = col - 1
+  end
+
+  local self = highlighter.active[buf]
+  if not self then
+    return {}
```
groner, line 31: @max397574 Should this be false or true or maybe nil? I noticed the example use you posted checks for true specifically, but I don't know what the empty table means.

Thanks for adding this, BTW.

smjonas: I agree with groner, what is `{}` supposed to mean? Shouldn't it be `false` which is what the documentation says? @max397574

viktomas:
Using the [Disabling completion in certain contexts, such as comments](https://github.com/hrsh7th/nvim-cmp/wiki/Advanced-techniques#disabling-completion-in-certain-contexts-such-as-comments) advanced setup broke Telescope for me (I couldn't use C-p and C-n to navigate the items).

My best guess is that the cmp plugin enablement is by default driven by this code:
https://github.com/hrsh7th/nvim-cmp/blob/5dce1b778b85c717f6614e3f4da45e9f19f54435/lua/cmp/config/default.lua#L10-L16

and that we overrode that with only checking for comments. That means that in the telescope prompt we enable the cmp plugin and that will take over the C-p and C-n keybindings.

I fixed it with 

```lua
enabled = function()
    local disabled = false
    disabled = disabled or (vim.api.nvim_buf_get_option(0, "buftype") == "prompt")
    disabled = disabled or (vim.fn.reg_recording() ~= "")
    disabled = disabled or (vim.fn.reg_executing() ~= "")
    disabled = disabled or context.in_treesitter_capture("string")
    return not disabled
end,
```

max397574:
yes you do overwrite the whole thing and you should do it like you did

b1nhack:
nvim_buf_get_option is deprecated.

```
enabled = function()
  local disabled = false
  disabled = disabled or (vim.api.nvim_get_option_value('buftype', { buf = 0 }) == 'prompt')
  disabled = disabled or (vim.fn.reg_recording() ~= '')
  disabled = disabled or (vim.fn.reg_executing() ~= '')
  disabled = disabled or require('cmp.config.context').in_treesitter_capture('comment')
  return not disabled
end,
```

PR #677, opened 2021-12-28
Add the snippet indicator for items with Snippet kind

dmitmel:
Completion results returned by https://github.com/dcampos/cmp-snippy are like this - that source defines an `execute` function which calls nvim-snippy directly, instead of letting nvim-cmp handle snippet expansion.


hrsh7th:
Looks good. Let's merge for the time being.

PR #690, opened 2021-12-30
handle `nil` value accessing `CompletionItemKind`

ap29600:
this prevents error messages from polluting the screen while using [ols](https://github.com/DanielGavin/ols) for autocompletion.


hrsh7th:
LGTM

PR #698, opened 2022-01-01
Handle nvim_{buf_}get_keymap return no rhs due to 'callback' mapping

### LINKED ISSUE #699: string.match error caused when using new keymap API callback
With the new `nvim_{buf_}{get,set}_keymap` API mappings can be defined "without" a right hand side (realistically it is just an empty string to avoid breaking the RPC API contract) and instead passing a `callback` to the `opts` argument.

The issue comes into play when we call `nvim_{buf_}get_keymap` which no longer returns a right hand side when the mapping instead defines a `callback`. Then when calling `string.match`, the right hand side of the mapping can be `nil` (if it is instead a `callback` mapping) so we get errors.

See #698  for a potential fix.

levouh:
This is the only issue I've run into when using the new `nvim_{buf_}set_keymap` functions added recently. Not sure if there is a cleaner way to fix it, but essentially we can no longer rely on the mappings having a `rhs` (as returned from `nvim_{buf_}get_keymap`) if they instead define a `callback` as part of the `opts` key.

These are the only place where I can see that this API function is used.

Fixes #699, original core _PR_ [here](https://github.com/neovim/neovim/commit/b411f436d3e2e8a902dbf879d00fc5ed0fc436d3)

### DIFF
--- a/lua/cmp/utils/keymap.lua
+++ b/lua/cmp/utils/keymap.lua
@@ -17,7 +17,7 @@ keymap.normalize = function(keys)
   vim.api.nvim_set_keymap('t', '<Plug>(cmp.utils.keymap.normalize)', keys, {})
   for _, map in ipairs(vim.api.nvim_get_keymap('t')) do
     if keymap.equals(map.lhs, '<Plug>(cmp.utils.keymap.normalize)') then
-      return map.rhs
+      return map.rhs or ""
     end
   end
   return keys
@@ -122,7 +122,7 @@ keymap.get_mapping = function(mode, lhs)
     if keymap.equals(map.lhs, lhs) then
       return {
         lhs = map.lhs,
-        rhs = map.rhs,
+        rhs = map.rhs or "",
         expr = map.expr == 1,
         noremap = map.noremap == 1,
         script = map.script == 1,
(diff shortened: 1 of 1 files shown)


hrsh7th:
So does we not be able to properly fallback if the callback function is set in nvim_buf_set_keymap?

levouh:
Not actually 100% sure, still tracing through how the fallback logic is handled.

Are you talking about [here](https://github.com/hrsh7th/nvim-cmp/blob/eca04af4f8881013498eb7e30aae3ea04c0e552a/lua/cmp/utils/keymap.lua#L89)?

arsham:
Thank you @levouh for your contribution. I've been having problems with this issue and I can confirm that your PR will fix it.
You rock!

levouh:
> Thank you @levouh for your contribution. I've been having problems with this issue and I can confirm that your PR will fix it. You rock!

It's a bit more of a "stop the stack trace" fix, but I'm not 100% sure it doesn't introduce other issues in the mean time. Requires a bit more feedback from @hrsh7th. 

Based on what I've looked at so far there seem to be some easier ways to do these things now that we can set/get Lua funcrefs for mappings but that is a much larger effort for sure.

hrsh7th:
I think we should support `nvim_get_keymap(...).callback`.
But I'll do it later.

PR #712, opened 2022-01-08
Default CmpItemKind%name% to CmpItemKind

isra17:
Reading the doc it seemed I could set `CmpItemKind` to override all item kind. Am I missing something?


hrsh7th:
it makes sense to me.

Issue #590, opened 2021-11-29
Tab keys are broken again


wookayin:
**Describe the bug**

In the command line, `<Tab>` and `<Shift-Tab>` keys are broken since ea10d5b, when cmdline support is not being used but some other plugins for commandline is being used.

![Screenshot](https://user-images.githubusercontent.com/1009873/143938596-f8a0a95c-5611-47e0-ad17-256665b8d1be.gif)

Please see the GIF. FYI I am using wilder.nvim.

1. Press Tab or Shift-Tab to invoke the completion
2. The completion menu does not appear or works as normally when pressing the keys; or weird characters are being inserted.

**To Reproduce**

Config

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
Plug 'gelguy/wilder.nvim'  " <--------------- NOTE: A new plugin that conflicts
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = {
    ['<CR>'] = cmp.mapping.confirm({ select = true })
  },

  sources = {
    { name = "nvim_lsp" },
    { name = "buffer" },
  },
}
EOF

lua << EOF
local capabilities = require('cmp_nvim_lsp').update_capabilities(vim.lsp.protocol.make_client_capabilities())

require'lspconfig'.cssls.setup {
  capabilities = capabilities,
}
EOF


" Wilder.nvim
augroup WilderLazyInit
    autocmd!
    autocmd CmdlineEnter * ++once call s:setup_wilder() | call wilder#main#start()
[...]

PR #720, opened 2022-01-10
Improve keymap handling

hrsh7th:
This is WIP. But I think we should always use `feedkeys` for solving the recursive keymapping.


wookayin:
As mentioned in #590, this solves evacuation behavior for `Tab/S-Tab`, but when I press <Up>/<Down> in the command mode I see raw key sequences like:
```
:e <Up><Up><Up><Up> 
```

This seems a minor side effect, so you may want to fix it before merging this.

hrsh7th:
@wookayin I think this `<Up>...` problem was solved now. It's easy miss. (I'd noticed it via fixing the tests).

hrsh7th:
Now. extra fallback key mapping is not needed anymore.

wookayin:
The `<Up>..` problem is gone but now `<Tab>` is broken again.

hrsh7th:
@wookayin Sorry. I've improved the implementation... maybe!

wookayin:
@hrsh7th This time everything works well! Awesome!

wookayin:
@hrsh7th, I know it's already late but I found that this actually made debugging keymappings very difficult.
```
c  <Plug>(cmp.u.k.fallback:9)   v:lua.cmp.utils.keymap.set_map(12)   
c  <Plug>(cmp.u.k.fallback:7)   v:lua.cmp.utils.keymap.set_map(9)
c  <Plug>(cmp.u.k.fallback:6) * v:lua.cmp.utils.keymap.set_map(7)
c  <Plug>(cmp.u.k.fallback:3) * v:lua.cmp.utils.keymap.set_map(3)
```

Is there any chance where we could have more human-interpretable idnetifiers as fallback/evacuate mappings?

hrsh7th:
Currently, there is a problem with the handling of key mapping.
Curing it is a top priority in nvim-cmp.

After the fix is complete, you can consider improving the debuggerability of the key mapping if you can afford it.

Google translated.

PR #727, opened 2022-01-13
Search mode horizontal view

tzachar:
When using `cmp-cmdline` in search mode (/ or ?), the custom entries menu is displayed vertically, which obscures the buffer when searching -- especially when using `incsearch`. This new view opens the entries in the statusline, similar to wilder.nvim, which solved the issue.

Some design choices:
1) Enabled by setting `experimental.horizontal_search = true`
2) Only the `abbr` field for each item is used
3) Items are separated by `|`
4) Fuzzy matches etc are highlighted using cmp highligh groups
5) the selected item is highlighted using `CursorLine`, as I don't think statusline supports blending hlgroups.

There might be some bugs, but I have been using it for a while.
Cheers.

### DIFF
--- a/lua/cmp/view/statusline_entries_view.lua
+++ b/lua/cmp/view/statusline_entries_view.lua
@@ -1,0 +1,259 @@
+local event = require('cmp.utils.event')
+local autocmd = require('cmp.utils.autocmd')
+local feedkeys = require('cmp.utils.feedkeys')
+local config = require('cmp.config')
+local types = require('cmp.types')
+local keymap = require('cmp.utils.keymap')
+local misc = require('cmp.utils.misc')
+local api = require('cmp.utils.api')
+
+---@class cmp.CustomEntriesView
+---@field private offset number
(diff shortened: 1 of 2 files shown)


hrsh7th:
Hm... It's interesting. But I think we must consider the configuration way. I'll consider it and play it.

lua/cmp/view/statusline_entries_view.lua
```diff
@@ -0,0 +1,259 @@
...
+
+statusline_entries_view.abort = function(self)
+  feedkeys.call('', 'n', function()
+    self:close()
+  end)
+end
+-- format entry for status line.
+-- only use the 'abbr' field
+-- return formatted text, original text, view length
+statusline_entries_view.highlight_entry = function(self, entry)
```
hrsh7th, line 102: shouldn't use statusline. you render items with floating window. I think.

tzachar:
Not sure how to do that and keep the pr

PR #729, opened 2022-01-13
Statusline view in search mode or command line mode

tzachar:
When using `cmp-cmdline` in search mode (/ or ?), the custom entries menu is displayed vertically, which obscures the buffer when searching -- especially when using `incsearch`. This new view opens the entries in the statusline, similar to wilder.nvim, which solved the issue.

Some design choices:
1) Enabled by setting `experimental.horizontal_search = true`
2) Only the `abbr` field for each item is used
3) Items are separated by `|`
4) Fuzzy matches etc are highlighted using cmp highligh groups
5) the selected item is highlighted using `CursorLine`, as I don't think statusline supports blending hlgroups.

There might be some bugs, but I have been using it for a while.
Cheers.

### DIFF
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -113,15 +113,17 @@ view.open = function(self, ctx, sources)
 
   -- complete_done.
   if #entries == 0 then
-    self.event:emit('complete_done', {
-      entry = self:_get_entries_view():get_selected_entry(),
-    })
     self:close()
   end
 end
 
 ---Close menu
 view.close = function(self)
+  if self:visible() then
+    self.event:emit('complete_done', {
+      entry = self:_get_entries_view():get_selected_entry(),
+    })
+  end
   self:_get_entries_view():close()
   self.docs_view:close()
   self.ghost_text_view:hide()
@@ -179,11 +181,11 @@ end
 ---Return current configured entries_view
 ---@return cmp.CustomEntriesView|cmp.NativeEntriesView
 view._get_entries_view = function(self)
-  local c = config.get()
   self.native_entries_view.event:clear()
   self.custom_entries_view.event:clear()
   self.statusline_entries_view.event:clear()
 
+  local c = config.get()
   if c.experimental.native_menu then
     self.native_entries_view.event:on('change', function()
       self:on_entry_change()
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -204,36 +204,33 @@ view._get_entries_view = function(self)
 end
 
 ---On entry change
-view.on_entry_change = async.throttle(
-  vim.schedule_wrap(function(self)
-    if not self:visible() then
-      return
+view.on_entry_change = async.throttle(function(self)
+  if not self:visible() then
+    return
+  end
+  local e = self:get_selected_entry()
+  if e then
+    for _, c in ipairs(config.get().confirmation.get_commit_characters(e:get_commit_characters())) do
+      keymap.listen('i', c, function(...)
+        self.event:emit('keymap', ...)
+      end)
     end
-    local e = self:get_selected_entry()
-    if e then
-      for _, c in ipairs(config.get().confirmation.get_commit_characters(e:get_commit_characters())) do
-        keymap.listen('i', c, function(...)
-          self.event:emit('keymap', ...)
-        end)
+    e:resolve(vim.schedule_wrap(self.resolve_dedup(function()
+      if not self:visible() then
+        return
       end
-      e:resolve(vim.schedule_wrap(self.resolve_dedup(function()
-        if not self:visible() then
-          return
-        end
-        self.docs_view:open(e, self:_get_entries_view():info())
-      end)))
-    else
-      self.docs_view:close()
-    end
+      self.docs_view:open(e, self:_get_entries_view():info())
+    end)))
+  else
+    self.docs_view:close()
+  end
 
-    e = e or self:get_first_entry()
-    if e then
-      self.ghost_text_view:show(e)
-    else
-      self.ghost_text_view:hide()
-    end
-  end),
-  20
-)
+  e = e or self:get_first_entry()
+  if e then
+    self.ghost_text_view:show(e)
+  else
+    self.ghost_text_view:hide()
+  end
+end, 20)
 
 return view
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -4,7 +4,7 @@ local event = require('cmp.utils.event')
 local keymap = require('cmp.utils.keymap')
 local docs_view = require('cmp.view.docs_view')
 local custom_entries_view = require('cmp.view.custom_entries_view')
-local statusline_entries_view = require('cmp.view.statusline_entries_view')
+local wildmenu_entries_view = require('cmp.view.wildmenu_entries_view')
 local native_entries_view = require('cmp.view.native_entries_view')
 local ghost_text_view = require('cmp.view.ghost_text_view')
 
@@ -13,6 +13,7 @@ local ghost_text_view = require('cmp.view.ghost_text_view')
 ---@field private resolve_dedup cmp.AsyncDedup
 ---@field private native_entries_view cmp.NativeEntriesView
 ---@field private custom_entries_view cmp.CustomEntriesView
+---@field private wildmenu_entries_view cmp.CustomEntriesView
 ---@field private change_dedup cmp.AsyncDedup
 ---@field private docs_view cmp.DocsView
 ---@field private ghost_text_view cmp.GhostTextView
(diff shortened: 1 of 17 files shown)


tzachar:
@hrsh7th 

Regarding not using the statusline: is there any benefit to using a float other than blending highlighting groups?

max397574:
@tzachar could you send a screenshot?

tzachar:
@max397574 
here you go
Peek 2022-01-14 13-30

tzachar:
@hrsh7th 
I moved the PR to using a floating window.

tzachar:
@hrsh7th 
Any chance of getting this merged?

tzachar:
@hrsh7th 
I've been using this PR for the past month. It really made a difference for me.
What more is required to get this merged? You mentioned something about configuration change, but did not elaborate.
Thanks.

Shougo:
Hm...  I think the feature should be configurable.
And the title should be `Statusline view in search mode or command line mode`.

hrsh7th:
I'll check it. Sorry for the late.

hrsh7th:
My random thoughts are here. I want to discuss these points.

##### The configuration option shouldn't be experimental

I think if we ship this feature, we don't plan to remove this. So it isn't experimental.
But I can't think reasonable option for this. Please let me your ideas.

##### It should be worked in insert-mode and normal cmdline-mode.

I don't think it's useful that it's worked in insert-mode. But it's consistent.

##### UI should be the same as native wildmenu.

But it's depending on to the personal preferences...

```lua
cmp.setup {
  ...
}

cmp.setup.cmdline('/', {
  ...
  some_reasonabole_option = true
  ...
})
```

hrsh7th:
BTW, It's useful :)

lua/cmp/view.lua
```diff
@@ -189,10 +192,14 @@ view._get_entries_view = function(self)
...
-    self.custom_entries_view.event:on('change', function()
+    local v = self.custom_entries_view
+    if vim.tbl_contains({ '/', '?' }, vim.fn.getcmdtype()) and c.experimental.horizontal_search then
+      v = self.statusline_entries_view
+    end
+    v.event:on('change', function()
       self:on_entry_change()
     end)
-    return self.custom_entries_view
+    return v
```
hrsh7th, line 202:
```lua
cmp.setup {
  appearance = {
    entries_view = 'custom' | 'statusline' | 'native', -- the `native` will be removed in the future.
  }
}
```

I'm afraid the `appearance` and `entries_view` are good wording or not. (I'm always afraid of such wording decisions...

hrsh7th:
BTW, It this behavior intented? It isn't the same as native wildmenu but not bad. It's interesting.

https://user-images.githubusercontent.com/629908/151798718-64c9b904-1201-4b16-aa30-a37dbbc3fc2b.mp4

lua/cmp/view.lua
gegoune: How about `display.layout` or `window.layout`?

tzachar:
which behaviour do you refer to?

tzachar:
> My random thoughts are here. I want to discuss these points.
> 
> ##### The configuration option shouldn't be experimental
> I think if we ship this feature, we don't plan to remove this. So it isn't experimental. But I can't think reasonable option for this. Please let me your ideas.
> 
> ##### It should be worked in insert-mode and normal cmdline-mode.
> I don't think it's useful that it's worked in insert-mode. But it's consistent.

If its not useful then why support it?

> ##### UI should be the same as native wildmenu.
> But it's depending on to the personal preferences...

Can you be more specific? 

> ```lua
> cmp.setup {
>   ...
> }
> 
> cmp.setup.cmdline('/', {
>   ...
>   some_reasonabole_option = true
>   ...
> })
> ```

lua/cmp/view.lua
hrsh7th: We already have `cmp.setup.cmdline`.
You can set appearance option for only cmdline-mode.

tzachar: Ok. So maybe we we can add the appearance config option both in general and in cmdline mode.
Makes sense?

hrsh7th: @tzachar Yes. I think this feature is needless in insert-mode. But it's useful to keep the consistency.

What do you think about this?
```
cmp.setup {
  view = {
    entries = 'custom' | 'statusline' | 'native',
  }
}
```

hrsh7th: Hm... `display` / `appearance` / `layout` / `window` / `view` / `ui`...

hrsh7th:
>> UI should be the same as native wildmenu.
>> But it's depending on to the personal preferences...

> Can you be more specific?

The following screencast uses `native wildmenu`.

The native wildmenu is

- Does not contains `|` chars.
- The items are separated by two spaces.
- The default highlight is `StatusLine` (maybe... I can't find actual one from vim help)
- The active item highlight is `WildMenu`.

https://user-images.githubusercontent.com/629908/151822440-03ace504-9d2c-4c3a-92fd-22cee5ad88af.mp4

hrsh7th:
Hm... But I think we shouldn't use wildmenu highlights. We should use pum highlights.

lua/cmp/view/statusline_entries_view.lua
```diff
@@ -0,0 +1,296 @@
...
+  self:init()
+
+  self.entries_win = window.new()
+  self.entries_win:option('conceallevel', 2)
+  self.entries_win:option('concealcursor', 'n')
+  self.entries_win:option('cursorlineopt', 'line')
+  self.entries_win:option('foldenable', false)
+  self.entries_win:option('wrap', false)
+  self.entries_win:option('scrolloff', 0)
+  self.entries_win:option('winhighlight', 'Normal:Pmenu,FloatBorder:Pmenu,CursorLine:PmenuSel,Search:None')
```
hrsh7th, line 42: We should create new highlight groups but it's ok at the moment.

lua/cmp/view/statusline_entries_view.lua
```diff
@@ -0,0 +1,296 @@
...
+end
+
+statusline_entries_view.abort = function(self)
+  feedkeys.call('', 'n', function()
+    self:close()
+  end)
+end
+
+statusline_entries_view.draw = function(self)
+  local entries_buf = self.entries_win:get_buffer()
```
hrsh7th, line 163: Probably, I'll try to refactor it using the `&sidescrolloff` and `vim.api.nvim_win_set_cursor` for managing viewport after merge this PR.

lua/cmp/view.lua
```diff
@@ -4,6 +4,7 @@ local event = require('cmp.utils.event')
 local keymap = require('cmp.utils.keymap')
 local docs_view = require('cmp.view.docs_view')
 local custom_entries_view = require('cmp.view.custom_entries_view')
+local statusline_entries_view = require('cmp.view.statusline_entries_view')
```
hrsh7th, line 7: I think the filename and varname shouldbe `wildmenu_entries_view`. (What do you think about it?)

lua/cmp/view.lua
hrsh7th: I think we can merge this after solve this.

lua/cmp/view.lua
tzachar: I dont really have an opinion here. I dont mind changing if you do.

lua/cmp/view.lua
hrsh7th: OK. I decided.
I'll organize the configuration options after merging this PR.

hrsh7th:
I will merge it once and organize it. Is that okay?

tzachar:
I think its ready to be commited now.
Config is like you suggested:
```lua
  view = {                  
    entries = 'custom'      
  }                         
````

and in cmdline:
```lua
cmp.setup.cmdline('/', {         
  view = {                       
    entries = 'wildmenu',        
    separator = '|',             
  },                            
}
```

hrsh7th:
I think the `separator` is needed only for `entries` and the value of `wildmenu`. So it's should be contained to the `{ name = 'wildmenu', separator = '|' }`

```lua
cmp.setup {
  view = {
    entries = { name = 'wildmenu', separator = '|' } -- the user can also specify the `wildmenu` literal string.
  }
}
```

tzachar:
> I think the `separator` is needed only for `entries` and the value of `wildmenu`. So it's should be contained to the `{ name = 'wildmenu', separator = '|' }`
> 
> ```lua
> cmp.setup {
>   view = {
>     entries = { name = 'wildmenu', separator = '|' } -- the user can also specify the `wildmenu` literal string.
>   }
> }
> ```
done

hrsh7th:
LGTM.

tzachar:
@hrsh7th 
On last thing I forgot: what about adding some documentation, pointing users to the new feature?

dsully:
Does this change address the issue of the "normal" vertical menu crossing over the status line?

tzachar:
probably. It overlays the status line with the entries.

dsully:
Hm. Sounds like a different issue then. I want the ("normal") vertical completion menu, but to have it render above the status line.

PR #753, opened 2022-01-23
Fix docs to avoid "E20: Mark not set" error

filipiz:
`autocmd` notation is invalid in the FAQ session.

The way it is set it generates an "E20: Mark not set" error when buffer of given filetype is opened.

This PR set correct `autocmd` notation.


hrsh7th:
LGTM.Thank you!

PR #777, opened 2022-02-05
Refactor context module, return false instead of {} in `context.in_treesitter_capture`

smjonas:
Not sure why returning `{}` was required (see discussion at https://github.com/hrsh7th/nvim-cmp/pull/676#discussion_r782408306).
Now we can omit the `== true` check in
```lua
enable = function()
    return not (
      context.in_treesitter_capture("comment") == true or context.in_syntax_group("Comment")
    )
end
```
-->

```lua
enable = function()
    return not (
      context.in_treesitter_capture("comment") or context.in_syntax_group("Comment")
    )
end
```
I also simplified the code at the end of the file.

### DIFF
--- a/lua/cmp/config/context.lua
+++ b/lua/cmp/config/context.lua
@@ -11,6 +11,7 @@ context.in_syntax_group = function(group)
       return true
     end
   end
+  return false
 end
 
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM

PR #781, opened 2022-02-10
Fix: close the view if cmp.enabled = false and the user types some text

smjonas:
A scenario where this lead to unintended behaviour previously was the following:

If cmp.enabled was configured such that it disabled cmp when in a comment,
and the user typed `--` in a lua file (this starts a comment), the cmp window
would not close if further text is typed on that line (although cmp should be disabled).

---

I have the following in my config:
```lua
enabled = function()
  local in_prompt = vim.api.nvim_buf_get_option(0, "buftype") == "prompt"
  if in_prompt then
    return false
  end
  return not(context.in_treesitter_capture("comment") == true or context.in_syntax_group("Comment"))
end,
```

To see what the issue is, have a look at the following GIF:
cmp_bug_pre

This PR should fix it:
cmp_bug_post

Let me know if there is a better way to fix this :)

### DIFF
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -309,12 +309,18 @@ end)
 autocmd.subscribe('TextChanged', function()
   if config.enabled() then
     cmp.core:on_change('TextChanged')
+  else
+    cmp.core:reset()
+    cmp.core.view:close()
   end
 end)
 
 autocmd.subscribe('CursorMoved', function()
   if config.enabled() then
     cmp.core:on_moved()
+  else
+    cmp.core:reset()
+    cmp.core.view:close()
   end
 end)
 


hrsh7th:
I think we should add it to CursorMoved too.

BTW, this solution is looks good to me.

hrsh7th:
I think both needed. :)

smjonas:
Alright, I have updated the PR

hrsh7th:
LGTM.

masakichi:
Hello, there.

After this PR got merged, seems the completion in search (when you type `/` in normal mode, then type any word)  is not functional now. When I comment out this PR's change(especially for `TextChanged` part), it works normally as before.

Shall we revert the change, at least for `TextChanged` part?

Thanks

smjonas:
@masakichi I have created #790 to revert the change to `TextChanged`. Since the addition to `CursorMoved` was not removed, it still fixes the original issue this PR addressed :) Sorry for introducing that bug, I am not using cmdline completion so didn't notice it ^^.

Issue #780, opened 2022-02-08
When used with nvim-autopairs, completion confirmed by typing '(' leads to two pairs of parenthesis being inserted


dmbfm:
This seems to happen because nvim-autopairs the `on_confirm_done` event. 

Let's stay I want to complete a function named `somefunc`.  I type `some`, go to the `somefunc` item in the completion menu, and then type the character `(`. What seems to happen is that the `on_confirm_done` event is called before the `(` character is inserted. Then nvim-autopairs does not see any parenthesis after the function name and inserts a pair `()`. Next, the parenthesis I type is inserted and paired by nvim-autopairs, and I end up with two pairs of parenthesis, i.e., `somefunc(())`.

Now, I want to be able to both complete a function by typing `(` or by using `<C-y>` or `<cr>`. 

As a tentative hack to fix this, I have added an option to suppress the `on_confirm_done` event on the `core.confirm` function:
https://github.com/dmbfm/nvim-cmp/blob/06b6471dce6783faa390a0519f53f603fce3d773/lua/cmp/core.lua#L432. The `core.key_map` function sets this to true when calling `core.confirm`.

I think this is pretty much a hack, and I haven't created a PR because probably this is not the "correct" way to fix this. 

Also, maybe this issue is too specific to my use case and most people have no need for such modifications.

yvesnrb:
I have the exact same problem! It's pretty annoying. I have now taken to using `<C-y>` to confirm completions, but I still slip up and hit the open parenthesis by accident on occasion, causing the double parenthesis problem. An option to suppress the confirm done event would be very nice indeed.

Shougo:
It seems nvim-autopairs problem instead of nvim-cmp.

hrsh7th:
I think the duplicate brackets were inserted by nvim-auopairs. What should I do in that situation?

dmbfm:
@hrsh7th Indeed it is inserted by nvim-autopairs... because they listen to the `on_confirm_done` event. Is there another way they could do this? I'm asking because I don't know that much about the internals of nvim-cmp, but I can try creating a PR for nvim-autopairs if I figure out a way better way to do this.

hrsh7th:
It seems possible to add whether core: confirm was called because of commitCharacter.

hrsh7th:
```
core:confirm({
  commit_character = nil or '(' or '.'
})
```

dmbfm:
Nice! I will try a solution on nvim-autopairs then. Thanks!

PR #782, opened 2022-02-10
Added `event_chars` property to `confirm_done event`

dmbfm:
Currently it seems there is no way for someone listening to the
`confirm_done` event to know exactly what character was used to commit
the completion confirmation.

This commit adds a `event_chars` property to the `confirm_done` event.
Its value is just the `chars` variable form the `core.on_keymap`
function.

This should help resolving a double parentheses issue with nvim-autopairs (#780).

### DIFF
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -119,6 +119,7 @@ core.on_keymap = function(self, keys, fallback)
     local is_printable = char.is_printable(string.byte(chars, 1))
     self:confirm(e, {
       behavior = is_printable and 'insert' or 'replace',
+      commit_chars = chars,
     }, function()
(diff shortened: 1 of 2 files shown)


lua/cmp/core.lua
```diff
@@ -119,6 +119,7 @@ core.on_keymap = function(self, keys, fallback)
     local is_printable = char.is_printable(string.byte(chars, 1))
     self:confirm(e, {
       behavior = is_printable and 'insert' or 'replace',
+      commit_chars = chars,
```
hrsh7th, line 122: could you change the name to commit_character?

dmbfm: Sure, no problem

PR #813, opened 2022-02-22
dev

hrsh7th:


lua/cmp/utils/window.lua
```diff
@@ -131,61 +131,59 @@ end
...
-      vim.api.nvim_win_set_option(self.thumb_win, 'winhighlight', 'EndOfBuffer:'..highlight..',NormalFloat:'..highlight)
+      vim.api.nvim_win_set_option(self.thumb_win, 'winhighlight', 'EndOfBuffer:'  .. highlight .. ',NormalFloat:'  .. highlight)
     end
-
     if self.scrollbar ~= '' then
-      local replace = {}
-      for i = 1, info.scrollbar.height do replace[i] = self.scrollbar end
-
-      vim.api.nvim_buf_set_lines(vim.api.nvim_win_get_buf(self.thumb_win), 0, -1, true, replace)
+      vim.api.nvim_buf_set_lines(vim.api.nvim_win_get_buf(self.sbar_win), 0, -1, true, misc.rep({ self.scrollbar }, style.height))
```
seblj, line 181: I am using the dev-branch since I really like the completion menu border feature. This line broke something for me, since now it errors almost on every keystroke. It's because `self.sbar_win` can be nil, and `vim.api.nvim_win_get_buf` expects a number. Don't really understand everything that happens here, but judging from the diff, should it be `self.thumb_win` here instead of `self.sbar_win`? It looks like it fixes it for me just testing really fast.

seblj: I just recreated the problem with a minimal config, and there is definitely an issue here. It looks like the issue is appearing when the documentation window needs to have a scrollbar. If I change it to `self.thumb_win` the issue disappears. See the videos I attached.

The following video is how it is right now. Notice how it doesn't error when it doesn't need to draw a scrollbar, but when it doesn't have enough space to show the entire documentation, an error appears.

https://user-images.githubusercontent.com/5160701/157007770-0007dd35-8bc7-41dc-af44-6eeb797e56a6.mov

The following video is how it is when I change `self.sbar_win` to `self.thumb_win` on line 181. Now it doesn't error and works just as expected

https://user-images.githubusercontent.com/5160701/157007858-3d800681-8869-47a7-b949-8b8b2e93d6f0.mov

I think one need to have a language server running so that there is a documentation window. If you want to try, simply install the lua sumneko language server, and use this minimal config.

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp', { 'branch': 'dev' }
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
    ['<C-d>'] = cmp.mapping(cmp.mapping.scroll_docs(4), { 'i', 'c' }),                                                                                         
    ['<C-u>'] = cmp.mapping(cmp.mapping.scroll_docs(-4), { 'i', 'c' }),                                                                                        
  },                                                                                                                                                           
                                                                                                                                                               
  window = {                                                                                                                                                   
    documentation = {                                                                                                                                          
      border = 'rounded',                                                                                                                                      
      scrollbar = '║',                                                                                                                                         
    },                                                                                                                                                         
                                                                                                                                                               
    completion = {                                                                                                                                             
      border = 'rounded',                                                                                                                                      
      scrollbar = '║',                                                                                                                                         
    },                                                                                                                                                         
  },                                                                                                                                                           
                                                                                                                                                               
    { name = "nvim_lsp" },                                                                                                                                     
    { name = "buffer" },                                                                                                                                       
  }),                                                                                                                                                          
}                                                                                                                                                              
                                                                                                                                                               
require('lspconfig').sumneko_lua.setup({})                                                                                                                     
EOF
```

@hrsh7th

hrsh7th: I think it's fixed. Thank you!

hrsh7th:
I have to consider the scrollbar appearance configuration.

lua/cmp/config.lua
```diff
@@ -165,6 +166,35 @@ config.normalize = function(c)
     c.mapping = normalized
   end
 
+  if type(c.window) == 'table' then
```
hrsh7th, line 169: recheck.

PR #836, opened 2022-03-12
Make sure config is not nil

nmattia:
This ensures that the config is at least `{}` and never `nil`.
Otherwise, calling `cmp.setup()` errors out.


hrsh7th:
seems good. Thank you.

Issue #231, opened 2021-09-21
Announcement: Breaking changes


hrsh7th:

hrsh7th:
#### Use floating window instead of native menu #224
- The user should change the mapping configuration.
    - Change `vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes('<C-n>', true, true, true, 'n', true)` to `cmp.select_next_item()`.
    - Change `vim.fn.pumvisible() == 1` to `cmp.visible()`

windwp:
Amazing i just test and it working like a charm. :)

Pocco81:
Did the config for `"<S-Tab>"` change as well?

hrsh7th:
Of cource, Yes.

Shatur:
I trying the following mapping, but nothing happens when I press `<C-Space>`:

```lua
    ['<C-Space>'] = function()
      if vim.fn.pumvisible() == 0 then
        cmp.mapping.complete()
      else
        cmp.mapping.confirm({ select = true })
      end
    end,
```

Am I configured something wrong?

wimstefan:
Can that be used already in conjunction with a snippet plugin?
How would the configuration look like?

Currently I'm using:
```lua
['<Tab>'] = function(fallback)
  if vim.fn.pumvisible() == 1 then
    vim.fn.feedkeys(vim.api.nvim_replace_termcodes('<C-n>', true, true, true), 'n')
  elseif vim.fn['snippy#can_expand_or_advance']() == 1 then
    vim.fn.feedkeys(vim.api.nvim_replace_termcodes('<Plug>(snippy-expand-or-next)', true, true, true), '')
  else
    fallback()
  end
end,
```

gegoune:
@hrsh7th Best to delete support post and lock the issue so only you can post here. This issue should be as clean as possible to allow users quickly pick up on breaking changes without support request interruptions. Do delete my post as well, of course.

@wimstefan no offence! :) (You only need to change two lines @hrsh7th posted, rest remains the same.)

vuki656:
Hey @hrsh7th could you please lock this thread and post here only if there are breaking changes so people can subscribe to get notifications? Thanks.

hrsh7th:
### SourceConfig.opts is deprecated https://github.com/hrsh7th/nvim-cmp/pull/561

```
cmp.setup {
  sources = {
    {
      name = 'buffer',
      opts = {
[...]

PR #844, opened 2022-03-16
Implement vim.api.nvim_create_autocmd

zer09:


smjonas:
This would increase the minimum supported version to nvim nightly / 0.7, so I think there should be a fallback to the old way of defining autocommands if an older version is used. Otherwise cmp would break for many users.

max397574:
@hrsh7th 0.7 is now released 🤷

hrsh7th:
Thank you ping.
I'll review ot.

max397574:
I think you can get rid of the `legacy` stuff since we can expect users to have 0.7 now (in about a week)

wookayin:
It'd be great if we could be careful about doing this, perhaps bumping up minimum version compatibility should take first (and pin the latest version compatible with neovim 0.6.x, etc.) preceding to this. Minimum neovim version requirement, is a breaking change, so it should be announced properly.

plugin/cmp.lua
```diff
@@ -3,59 +3,39 @@ if vim.g.loaded_cmp then
...
 vim.g.loaded_cmp = true
 
-local api = require "cmp.utils.api"
+local api = require('cmp.utils.api')
 local misc = require('cmp.utils.misc')
 local types = require('cmp.types')
 local config = require('cmp.config')
 local highlight = require('cmp.utils.highlight')
+local emit = require('cmp.utils.autocmd').emit
+local autos = require('cmp.autos')
```
wookayin, line 12: Please consider having a less confusing name, e.g., `cmp.autos` -> `cmp.autocmds`

hrsh7th:
> It'd be great if we could be careful about doing this, perhaps bumping up minimum version compatibility should take first (and pin the latest version compatible with neovim 0.6.x, etc.) preceding to this. Minimum neovim version requirement, is a breaking change, so it should be announced properly.

Of course, I'll add the announcement to the https://github.com/hrsh7th/nvim-cmp/issues/231 .
Isn't this enough for you?

hrsh7th:
I'd choose the way of announcement breaking changes via post it as planning status on https://github.com/hrsh7th/nvim-cmp/issues/231#issuecomment-1102917158

mehalter:
Should we also adapt to use `vim.api.nvim_set_hl` for reacting highlight links as well as `vim.api.nvim_create_user_command` for commands like `:CmpStatus`?

gegoune (changes requested):

plugin/cmp.lua
```diff
@@ -9,7 +9,7 @@ local types = require('cmp.types')
 local config = require('cmp.config')
 local highlight = require('cmp.utils.highlight')
 local emit = require('cmp.utils.autocmd').emit
-local autos = require('cmp.autos')
+local autos = require('cmp.autocmds')
```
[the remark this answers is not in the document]
zer09: My bad, I forgot to changed it

lua/cmp/autocmds.lua
```diff
@@ -0,0 +1,130 @@
...
+    callback = _G.cmp.plugin.cmdline.enter,
+  })
+
+  vim.api.nvim_create_autocmd('CmdwinEnter', {
+    group = ___cmp___,
+    callback = _G.cmp.plugin.cmdline.leave,
+  })
+end
+
+local function legacy_auto()
```
hrsh7th, line 72: Sorry but the nvim0.u is released so we don't need fallback to old implementation.
Please use only vim.api.*

lua/cmp/autocmds.lua
```diff
@@ -0,0 +1,130 @@
+local M = {}
+
+local v = vim.version()
+local has_native_autocmd = v.major >= 0 and v.minor >= 7
```
wookayin, line 4: I think this is going to be removed because we don't need fallback, but this can go wrong if neovim version becomes v1.0 (v.major == 1 and v.minor == 0)

wookayin:
One issue of using the new `nvim_create_autocmd` APIs is that it can make debugging quite hard, as the content of Lua callback functions are not shown.

See:
- https://www.reddit.com/r/neovim/comments/u5wpfl/how_to_know_the_details_debugget_location_of_an/
- https://github.com/hrsh7th/nvim-cmp/pull/922#discussion_r853449984
- https://github.com/neovim/neovim/issues/18153

Compare the output of `:autocmd`: the current, vim.cmd way would be easier to read

```
--- Autocommands ---
___cmp___  CmdWinEnter
    *         call v:lua.cmp.plugin.cmdline.leave() " for entering cmdwin with `<C-f>`
___cmp___  CmdlineEnter
    *         call v:lua.cmp.plugin.cmdline.enter()
```

than when `nvim_create_autocmd` is used:
```
___cmp___  CmdWinEnter
    *         <Lua function 500>
___cmp___  CmdlineEnter
    *         <Lua function 500>
```

So it seems that **we should use the field `desc`** so that one can easily locate where the callback functions are defined (actually one may search `autocmds.lua` or for `nvim_create_autocmd` though). There is another bug related to this, see 
https://github.com/neovim/neovim/issues/17588 and https://github.com/neovim/neovim/issues/18153.

zer09:
I just used the old `autocmd` for the `desc`. About this bug https://github.com/neovim/neovim/issues/17588, I can only just test on `5.16.18-1-MANJARO` so far it works fine.

lua/cmp/autocmds.lua
```diff
@@ -1,23 +1,22 @@
...
 
-local function nvim_auto()
+function M.autocmd()
   local ___cmp___ = vim.api.nvim_create_augroup('___cmp___', { clear = true })
   vim.api.nvim_create_autocmd('InsertEnter', {
     group = ___cmp___,
     callback = function()
       emit('InsertEnter')
     end,
+    desc = [[autocmd InsertEnter * lua require'cmp.utils.autocmd'.emit('InsertEnter')]],
```
gegoune, line 11: I would recommend making those descriptions human readable rather than insinuating they look like viml version.

`desc = 'nvim-cmp: emits InsertEnter event(?)',`

hrsh7th:
Thank you for your contribution!

I'll merge this at the moment and I'll refactor a bit on the `dev` branch.

hrsh7th:
LGTM

PR #848, opened 2022-03-18
Bottom up mode for custom menu

tzachar:
When in command line mode, the custom entry window opens up to an
unexpected height, depending on the current count of completion items.
The above makes it hard to anticipate where to look at, and makes life a bit
harder.

This patch adds an option to open the custom entries view in a bottom up
mode, and flips the regular behaviour of next/prev entry in this mode.
Setup is as easy as:

```
cmp.setup.cmdline(':', {
  view = {
    entries = {name = 'custom', selection_order = 'near_cursor' } 
  }
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ lua <<EOF
   -- Set configuration for specific filetype.
   cmp.setup.filetype('gitcommit', {
     sources = cmp.config.sources({
-      { name = 'cmp_git' }, -- You can specify the `cmp_git` source if you were installed it. 
+      { name = 'cmp_git' }, -- You can specify the `cmp_git` source if you were installed it.
     }, {
       { name = 'buffer' },
     })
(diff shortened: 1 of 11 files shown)


tzachar:
@hrsh7th 
Any chance of merging this?
10x

hrsh7th:
I think this must be smarter because the current option is only worth cmdline-completion.

For example, we can determine the direction via `the window is upper the cursor or not`.

I don't imagine the suitable option name but it should be changed, I think. e.g.: `selection_order = 'top_down' | 'near_cursor'`.

tzachar:
I like your idea.
Will try to implement.

hrsh7th:
Please consider the suitable option name :)

tzachar:
sure

tzachar:
@hrsh7th 
I think this last commit fixes all issues.
If you have nothing else, I would appreciate if you can merge this.
Cheers.

max397574:
Could you share an image or a gif of how this will look like?

tzachar:
Peek 2022-03-30 15-05

tzachar:
@hrsh7th 
Apologies if im pushing, but this has been ready for a while now.
Do you intend to merge?

hrsh7th:
LGTM.
I've checked the behavior. I don't like this behavior but someone might like this may be.

tzachar:
Thank you.
Ill add a wiki for this feature.

rignet-bicteg:
This PR broke command line autocompletion from Vim proper for me. When I open vim and do :e <Tab> I get an `I^`.

hrsh7th:
I can't reproduce. How do you determine the commit that broke the behavior?

tzachar:
I think this is something different. nvim now differentiats between <c-tab>
and <c-I>.

cinuor:
same to me, Tab in command line output like this
image

Tab and S-Tab mapped by cmp，but i do not set those in my cmp mapping config

hrsh7th:
This is not related to this PR. I've fixed it in latest main.

PR #857, opened 2022-03-27
use only the first filetype for formatting

ZwodahS:
This fix syntax highlighting when handling files with multiple-filetype.

Screenshot 2022-03-27 at 2 57 00 PM

After fix

Screenshot 2022-03-27 at 2 58 07 PM

If there is a better way to fix this, let me know :)


hrsh7th:
LGTM

Issue #485, opened 2021-11-06
cmdline completion affecting incsearch highlights


axieax:
**Describe the bug**
Incsearch highlights flicker when cmdline buffer completion is enabled. It seems that when no completion item is selected, the highlights will be quickly turned off, causing the "flickering" effect when typing.

**Minimal config based on [this](https://github.com/hrsh7th/nvim-cmp/blob/main/utils/vimrc.vim)**

```lua
vim.o.incsearch = true
vim.o.completeopt = "menu,menuone,noselect"
cmp.setup.cmdline("/", {
  sources = {
    { name = "buffer" },
  },
})
```
[See more](https://github.com/axieax/dotconfig/blob/main/nvim/lua/lsp/cmp.lua)

**To Reproduce**

1. Try searching the buffer with my settings
2. Highlights will go crazy

https://user-images.githubusercontent.com/62098008/140614408-69942d2b-df51-4f04-9423-6fde44e54b2e.mp4

**Expected behavior**

Highlights will persist even if no completion item is selected - using whatever is being typed.

**Additional context**

hrsh7th:
I think it can't be fixed because wilder.nvim has a same problem. (I think wilder.nvim is more detailed with cmdline completion, so I think it would be difficult if wilder.nvim couldn't.

PR #859, opened 2022-03-28
fix(misc): kill redundant redraw during incsearch (#485)

kevinhwang91:
I just use nvim-cmp to test nvimlsp and found the redraw issue for incsearch.

`<C-r><Esc>` is less side effect than `<C-r>=""<CR>`


hrsh7th:
LGTM. It works great. Thank you!

PR #886, opened 2022-04-11
fix: stop popup buffers from staying in memory after they close

bbbert:
I noticed that popup menu buffers stick around in the buffer list even after they are hidden, at least for a while. You can see them using `:ls!`. From the vim docs:

https://github.com/neovim/neovim/blob/1186f7dd96b054d6a653685089fc845a8f5d2f27/runtime/doc/windows.txt#L1274-L1282

This manifests in nvim-cmp's popups appearing in the buffer list in [Airline](https://github.com/vim-airline/vim-airline), and possibly other statusline plugins too.

Screenshot_2022-04-10_22-31-42

Changing to 'wipe' deletes the buffer immediately after it is hidden deletes the scratch buffers immediately and prevents this, and doesn't seem to affect functionality.

https://github.com/neovim/neovim/blob/1186f7dd96b054d6a653685089fc845a8f5d2f27/runtime/doc/options.txt#L1068-L1070

### DIFF
--- a/lua/cmp/utils/buffer.lua
+++ b/lua/cmp/utils/buffer.lua
@@ -21,7 +21,7 @@ buffer.ensure = function(name)
     created_new = true
     buf = vim.api.nvim_create_buf(false, true)
     vim.api.nvim_buf_set_option(buf, 'buftype', 'nofile')
-    vim.api.nvim_buf_set_option(buf, 'bufhidden', 'hide')
+    vim.api.nvim_buf_set_option(buf, 'bufhidden', 'wipe')
     buffer.cache[name] = buf
   end
   return buf, created_new


hrsh7th:
I don't know the difference in cost between reusing a single buffer and creating a scratch buffer for each render.

bbbert:
Hmm, yeah I don't know either. By the way, I just want to clarify that I am relatively new to neovim internals/neovim plugin development. I just observed this issue with the buffer listing in Airline and the buffer lists and this was the way I found that resolved it. I'm not entirely sure that this is the most correct way of handling buffers either. Happy to learn more and investigate further with you, and thanks for creating the plugin in the first place!

A little more digging:

Neovim seems to allocate a new buffer regardless and won't reuse old scratch buffers that are not wiped:
https://github.com/neovim/neovim/blob/cdfb045ea027e018f14fb2735c3fd6aaa718ae58/src/nvim/api/vim.c#L1052

Other plugins that make popup windows also seem to set `bufhidden=wipe`:
https://github.com/echasnovski/mini.nvim/blob/e6fdc2b39191245bc0751c55341475082652c380/lua/mini/starter.lua#L1232
https://github.com/ms-jpq/coq_nvim/blob/e00f88f4557c80b367345d323d1e46b885f9b7f8/coq/server/registrants/preview.py#L224

The reason why this worked in my case to hide the popup buffers from airline is because Airline checks both `buftype=nofile` and `bufhidden=wipe` to decide to ignore the buffer in the list:
https://github.com/vim-airline/vim-airline/blob/be5bda1f1dbfa4e36b2adabaf3f423adfa66c336/autoload/airline/extensions/tabline/buflist.vim#L34

tomtomjhj:
Airline shows the buffer only if [`buflisted()` is true](https://github.com/vim-airline/vim-airline/blob/be5bda1f1dbfa4e36b2adabaf3f423adfa66c336/autoload/airline/extensions/tabline/buflist.vim#L55), and cmp's popup buffer is unlisted (`nvim_create_buf(false, ..)`). So cmp's popup buffers won't appear in airline. @bbbert Can you double-check?

Also
> Neovim seems to allocate a new buffer regardless and won't reuse old scratch buffers that are not wiped:

IIUC this is not really an issue here, because cmp calls `nvim_create_buf` only if there's no corresponding buffer. With this patch, `nvim_create_buf` will be called everytime the popup is invoked.

tomtomjhj:
btw setting `buftype` (and `bufhidden` before this patch) looks redundant because `nvim_create_buf(.., true)` already does it:
https://github.com/neovim/neovim/blob/cdfb045ea027e018f14fb2735c3fd6aaa718ae58/src/nvim/api/vim.c#L1073-L1074

bbbert:
@tomtomjhj you're right about the buffer caching and the options being redundant. As for airline, looks like that check changed [8 days](https://github.com/vim-airline/vim-airline/commit/35af3bf37f41ec2fbd68cdb9176db02811869fbf) ago and got reverted back to buflisted [yesterday](https://github.com/vim-airline/vim-airline/pull/2526/files) ago, and I got caught right in the middle.

So the flag here turns out to *not* to be the cause of the issue I saw. Sorry, my bad. :/

The merits of this flag change, then, are a matter of performance. I'm not sure how to evaluate this. The time spent creating a new buffer seems quite small and my guess is that it doesn't matter that much either way. Thoughts?

PR #900, opened 2022-04-14
Revert #886 and refactor

tomtomjhj:
As discussed in https://github.com/hrsh7th/nvim-cmp/pull/886#issuecomment-1098596708, the issue @bbbert was experiencing has been fixed in airline. So I think it's ok to revert #886.


hrsh7th:
Thank you~!

Issue #846, opened 2022-03-18
[Request] Customizing aspects of custom_entries_view.lua


CKolkey:
I've been working on customizing some visual aspects of the custom popup menu, and ran into some issues that required making minor changes in `custom_entries_view.lua`. I don't mind this, but since I may not be the only one who is interested in this level of tweaking, I'm curious if there might be some possibility to expose these variables in the setup function. Alternatively, if there's a better way to accomplish what I'm trying to do, I'm all ears.

Specifically, I'm looking to be able to change:
- `custom_entries_view.lua:11`: `local SIDE_PADDING` to `0` so there is no window to the left of the 'kind' icon.
- `custom_entries_view.lua:35`: remove `CursorLine:PmenuSel` from `winhighlight` options so the background color of the 'kind' icon isn't changed by `PmenuSel`
- `custom_entries_view.lua:168`: offset `col` value by `-3` (`col = math.max(0, col - 3)`) so the completion entries line up with the text in the buffer.

This is how it looks without those changes:
![Screenshot 2022-03-18 at 10 46 00](https://user-images.githubusercontent.com/7228095/158980826-2ce2aa0e-4c92-408d-b83e-57fd34026e48.png)

And how it looks with the changes:
![Screenshot 2022-03-18 at 10 46 49](https://user-images.githubusercontent.com/7228095/158980870-5410592e-1480-4b3f-8d1c-c7756e8e9147.png)

Thanks in advance :)

hrsh7th:
I can understand your expectation.

But I can't imagine we achive it in general way. we might not derrived the proper abstraction...

The nvim-cmp's design already confusing. For example, we have `window` and `view` configuration section.

We must re-thinking the design to customize appearance.

hrsh7th:
I think the custom entries view is now customizable.

wookayin:
@hrsh7th How? Can you please let us know relevant commits/PRs or documentation?

hrsh7th:
See docs https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt#L624

PR #901, opened 2022-04-14
Add Completion Window Options

CKolkey:
Hi! Previously I opened #846 to see about exposing some configuration options of the completion window to users. With some of the recent changes you made, this became very simple to implement. Let me know if this is mergeable, or if I need to add anything else.

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -527,6 +527,16 @@ window.{completion,documentation}.zindex~
   `number`
   The completion window's zindex.
 
+                                       *cmp-config.window.completion.col_offset*
+window.completion.col_offset~
+  `number`
+  Offsets the completion window relative to the cursor.
+
+                                     *cmp-config.window.completion.side_padding*
+window.completion.side_padding~
+  `number`
+  The ammount of padding to add on the completion window's sides
+
                                      *cmp-config.window.documentation.max_width*
 window.documentation.max_width~
   `number`
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -85,6 +85,8 @@ return function()
       completion = {
         border = { '', '', '', '', '', '', '', '' },
         winhighlight = 'Normal:Pmenu,FloatBorder:Pmenu,CursorLine:PmenuSel,Search:None',
+        col_offset = 0,
+        side_padding = 1,
       },
       documentation = {
         max_height = math.floor(WIDE_HEIGHT * (WIDE_HEIGHT / vim.o.lines)),
--- a/lua/cmp/config/window.lua
+++ b/lua/cmp/config/window.lua
@@ -6,6 +6,8 @@ window.bordered = function(opts)
     border = opts.border or 'rounded',
     winhighlight = opts.winhighlight or 'Normal:Normal,FloatBorder:Normal,CursorLine:Visual,Search:None',
     zindex = opts.zindex or 1001,
+    col_offset = opts.col_offset or 0,
+    side_padding = opts.side_padding or 1
   }
 end
 
--- a/lua/cmp/view/custom_entries_view.lua
+++ b/lua/cmp/view/custom_entries_view.lua
@@ -8,8 +8,6 @@ local keymap = require('cmp.utils.keymap')
 local misc = require('cmp.utils.misc')
 local api = require('cmp.utils.api')
 
-local SIDE_PADDING = 1
-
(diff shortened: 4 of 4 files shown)


doc/cmp.txt
```diff
@@ -527,6 +527,16 @@ window.{completion,documentation}.zindex~
   `number`
   The completion window's zindex.
 
+                                       *cmp-config.window.completion.col_offset*
+window.completion.col_offset~
+  `number`
+  Offsets the completion window relative to the cursor.
```
gegoune, line 533: What if cursor is on column lesser than `col_offset`? Perhaps it should take a function and one co.uld calculate offset taking cursor position (could make function receive it as an argument to make it easier) into consideration

CKolkey: With a `col_offset` of `-10`, and the cursor at the first column of a buffer, it won't go over the edge of the window:
Screenshot 2022-04-14 at 10 33 00

However, if there's a split, it'll be offset by the amount specified:
Screenshot 2022-04-14 at 10 35 34


IMO this is what I would expect.

gegoune: Ah, ok then. Assumed window could go behind visible viewport. Thanks for clarifying. I am wondering whether there is a use case in which function (you could always accept either integer or function) would still be useful.

CKolkey: I suppose it's possible, though not for my use case :) I just want to have icons on the left and line the text up nicely

par4m:
excatly what i needed, is it possible to add this for Documentation window (relative to Completion window, not cursor) as well as cmdline ?

hrsh7th:
I think we should consider the configuration option.

I think the following API's is more flexible.

```lua
cmp.setup {
  window = {
    documentation = {
      on_open_before = function(...)
      end
    }
  }
}
```

I'm not happy with the naming and interface above, but I think it's better than the current one.

CKolkey:
I'm not sure I understand what you mean. Since there is already a declarative configuration API (`window.completion.border`, `window.completion.winhighlight`, etc), I think it would make more sense to continue in the same style. 

That being said, it's your project, so I'll be happy to implement it how you like once I understand what that is :)

hrsh7th:
I think we should re-design the way to customize window appearance in the future.
But this PR is enough maintenable and worth it.

lua/cmp/config/default.lua
```diff
@@ -85,6 +85,8 @@ return function()
       completion = {
         border = { '', '', '', '', '', '', '', '' },
         winhighlight = 'Normal:Pmenu,FloatBorder:Pmenu,CursorLine:PmenuSel,Search:None',
+        col_offset = 0,
+        side_padding = 1,
       },
```
hrsh7th, line 90: I think we also add the `col_offset` and `side_padding` support for `config/window.lua`.

CKolkey: I've added the fields and the ability to overwrite them via the table passed in (`opts`). Is there any more to it than that?

hrsh7th:
LGTM. Thank you~!

fitrh:
Sorry to comment out here, I don't know if this is a problem or something to be expected, but did you @CKolkey realize that if we set `side_padding` to 0, we get an extra space at the end of the item?

20220622_221731_full

Doing some quick patch like this

```diff
diff --git a/lua/cmp/view/custom_entries_view.lua b/lua/cmp/view/custom_entries_view.lua
index d3e97c9..a627f7c 100644
--- a/lua/cmp/view/custom_entries_view.lua
+++ b/lua/cmp/view/custom_entries_view.lua
@@ -143,7 +143,7 @@ custom_entries_view.open = function(self, offset, entries)
   vim.api.nvim_buf_set_option(entries_buf, 'modified', false)
 
   local width = 0
-  width = width + 1
+  width = width + config.get().window.completion.side_padding
   width = width + self.column_width.abbr + (self.column_width.kind > 0 and 1 or 0)
   width = width + self.column_width.kind + (self.column_width.menu > 0 and 1 or 0)
   width = width + self.column_width.menu + 1
```

Fix that behavior

20220622_222041_full

But since it's just a quick patch and I don't dig deep enough into the rendering stuff, I don't think it's the right solution.

CKolkey:
No, I didn't realize. I think you could likely use the `side_padding` when calculating window width, and that would solve it.

max397574:
@fitrh where can we find the config for that amazing looking cmp?

PR #915, opened 2022-04-18
Update README.md with reasonable defaults

YuvalHelman:
As the default for traversing the Autocomplete list was recently removed, having these defaults as the recommended configuration for new users will greatly reduce redundant questions from these new users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,8 @@ lua <<EOF
       -- documentation = cmp.config.window.bordered(),
     },
     mapping = cmp.mapping.preset.insert({
+      ['<C-p>'] = cmp.mapping.select_prev_item(),
+      ['<C-n>'] = cmp.mapping.select_next_item(),
       ['<C-b>'] = cmp.mapping.scroll_docs(-4),
       ['<C-f>'] = cmp.mapping.scroll_docs(4),
       ['<C-Space>'] = cmp.mapping.complete(),


hrsh7th:
I think this is not needed.

The default `cmp.mapping.preset.insert` defined as the following.

```lua
    ['<C-n>'] = {
      i = mapping.select_next_item({ behavior = types.cmp.SelectBehavior.Insert }),
    },
    ['<C-p>'] = {
      i = mapping.select_prev_item({ behavior = types.cmp.SelectBehavior.Insert }),
    },
```

YuvalHelman:
Does it work for you out of the box? because it doesn't for me. 
without adding the configuration I stated, I can't traverse the completion list.

hrsh7th:
@YuvalHelman Yes. It's worked out of the box.

```
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/vim-vsnip'
call plug#end()
PlugInstall | quit

lua << EOF
  local cmp = require "cmp"
  cmp.setup {
    snippet = {
      expand = function(args)
      end,
    },
    mapping = cmp.mapping.preset.insert(),
    sources = {
      { name = "buffer" },
    }
  }
EOF
```

PR #919, opened 2022-04-19
fix: use vim.treesitter.query instead of deprecated nvim-treesitter.t…

ecosse3:
…s_utils

Using vim.treesitter.query instead of deprecated nvim-treesitter.ts_query in https://github.com/nvim-treesitter/nvim-treesitter/commit/c3848e713a8272e524a7eabe9eb0897cf2d6932e

Fixes:
image

### DIFF
--- a/lua/cmp/config/compare.lua
+++ b/lua/cmp/config/compare.lua
@@ -175,7 +175,6 @@ compare.scopes = setmetatable({
     if ok then
       local win, buf = vim.api.nvim_get_current_win(), vim.api.nvim_get_current_buf()
       local cursor_row = vim.api.nvim_win_get_cursor(win)[1] - 1
-      local ts_utils = require('nvim-treesitter.ts_utils')
 
       -- Cursor scope.
       local cursor_scope = nil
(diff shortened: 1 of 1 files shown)


lua/cmp/config/compare.lua
```diff
@@ -205,7 +205,7 @@ compare.scopes = setmetatable({
         for _, definition in pairs(definitions) do
           if s <= definition.node:start() and definition.node:end_() <= e then
             if scope:id() == locals.containing_scope(definition.node, buf):id() then
-              local text = ts_utils.get_node_text(definition.node)[1]
+              local text = ts_query.get_node_text(definition.node)[1]
```
hrsh7th, line 208: Please use `vim.treesitter.query.get_node_text` directly here.

hrsh7th:
Thank you for contribution!

hrsh7th:
LGTM. Thank you~!

wookayin:
It seems that if we are to get the first line only, both function behaves the same.

wookayin:
f9f6d8dc49d5ae8c92cad79974b754e7e7f0417b

PR #922, opened 2022-04-19
Implement nvim_create_user_command

mehalter:
This moves the creation of the `:CmpStatus` command to the new `vim.api.nvim_create_user_command` API

This requires neovim 0.7 so it should be done along with #844

### DIFF
--- a/plugin/cmp.lua
+++ b/plugin/cmp.lua
@@ -102,7 +102,9 @@ if vim.on_key then
   end, vim.api.nvim_create_namespace('cmp.plugin'))
 end
 
-vim.cmd([[command! CmpStatus lua require('cmp').status()]])
+vim.api.nvim_create_user_command('CmpStatus', function()
+  require('cmp').status()
+end, { desc = 'Check status of cmp sources' })
 
 vim.cmd([[doautocmd <nomodeline> User CmpReady]])
 


plugin/cmp.lua
```diff
@@ -122,7 +122,7 @@ if vim.on_key then
   end, vim.api.nvim_create_namespace('cmp.plugin'))
 end
 
-vim.cmd [[command! CmpStatus lua require('cmp').status()]]
+vim.api.nvim_create_user_command('CmpStatus', require('cmp').status, {})
```
wookayin, line 125: Thanks for helping migrate to neovim 0.7.x APIs, but one downside of lua-based command definition is it will make debugging a bit hard. You can try `:Telescope commands` to see its definition reads like:
```
CmpStatus         0                        <Lua function 78>          
```
Meh, what is `<Lua function 78>`? So I would recommend adding `desc = ...` in the opts table.

mehalter: Thanks for this! I didn't even think about that. I just updated my commit to include a description. Let me know if you want me to change the wording of it!

plugin/cmp.lua
```diff
@@ -122,7 +122,11 @@ if vim.on_key then
   end, vim.api.nvim_create_namespace('cmp.plugin'))
 end
 
-vim.cmd [[command! CmpStatus lua require('cmp').status()]]
+vim.api.nvim_create_user_command(
+  'CmpStatus',
+  require('cmp').status,
```
gegoune, line 127: Please be aware that it will require cmp at the time that command is created. To avoid that it should be wrapped in lambda/anonymous function.

mehalter: I was thinking this as well. By the time this file is loaded doesn't that imply that `cmp` is available at the time? Is there a case where this `plugin/cmp.lua` file will be loaded but the plugin itself will not?

mehalter: I tested this `nvim_create_user_command` with some pretty heavy lazy loading in different cases and couldn't get it in a state where this caused issues when the `vim.api.nvim_create_user_command` was called.

gegoune: True, but `plugin/cmp.lua` will be sourced by nvim as a startup script (if added to rtp). So those requires are happening any way https://github.com/mehalter/nvim-cmp/blob/command_api/plugin/cmp.lua#L6-L10.

mehalter: I see, so just to clarify, should I update it to wrap it in an anonymous function?

gegoune: @hrsh7th Have you got an opinion here?

hrsh7th: For my personal taste, I think it's better to wrap it in an anonymous function. (This is because we can eliminate the assumption that the arguments given to the anonymous function implicitly match the status function.)

_google translated_

mehalter: This is a great point @hrsh7th I have updated the commit to wrap it in an anonymous function.

hrsh7th:
LGTM. Thank you!

PR #923, opened 2022-04-19
fix: Fix stylua errors and install_stylua script

wookayin:
The CI/Actions has been broken for a while (we shouldn't be ignoring build errors) and this PR should fix the CI build errors due to stylua.

- Fix code style issues, suggested by stylua
- Update install_stylua.lua script so that it can run in macOS as well.


hrsh7th:
Thank you!

PR #925, opened 2022-04-20
Move highlights to nvim_set_hl lua API

mehalter:
This gives a first stab at simplifying the highlighting in `cmp` to use the new `vim.api.nvim_set_hl` lua API.

I did remove the `cterm` and `gui` specific highlights because the `vim.fn.synIDattr` and `vim.api.nvim_set_hl` commands default to the current mode of neovim (i.e. gui/cterm/term) so I don't think it is necessary that we set all of them, instead just get the current mode. To my knowledge this mode won't change in the middle of running neovim.

This requires neovim 0.7 so it should be done along with https://github.com/hrsh7th/nvim-cmp/pull/844

### DIFF
--- a/lua/cmp/utils/highlight.lua
+++ b/lua/cmp/utils/highlight.lua
@@ -1,46 +1,30 @@
 local highlight = {}
 
 highlight.keys = {
-  'gui',
-  'guifg',
-  'guibg',
-  'cterm',
-  'ctermfg',
-  'ctermbg',
+  'fg',
+  'bg',
+  'bold',
+  'italic',
+  'reverse',
+  'standout',
+  'underline',
+  'undercurl',
+  'strikethrough',
 }
 
-highlight.inherit = function(name, source, override)
-  local cmd = ('highlight default %s'):format(name)
+highlight.inherit = function(name, source, settings)
   for _, key in ipairs(highlight.keys) do
-    if override[key] then
-      cmd = cmd .. (' %s=%s'):format(key, override[key])
-    else
-      local v = highlight.get(source, key)
-      v = v == '' and 'NONE' or v
-      cmd = cmd .. (' %s=%s'):format(key, v)
-    end
-  end
-  vim.cmd(cmd)
-end
-
-highlight.get = function(source, key)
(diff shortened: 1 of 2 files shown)


lua/cmp/utils/highlight.lua
```diff
@@ -1,46 +1,30 @@
...
-  'guifg',
-  'guibg',
-  'cterm',
-  'ctermfg',
-  'ctermbg',
+  'fg',
+  'bg',
+  'bold',
+  'italic',
+  'reverse',
```
gegoune, line 8: Isn't it `inverse` for that API?

mehalter: For that API it is actually `reverse`. I found this yesterday when `nightfox.nvim` moved to using the `nvim_set_hl` API: https://github.com/EdenEast/nightfox.nvim/issues/136

plugin/cmp.lua
```diff
@@ -47,65 +47,44 @@ end)
...
 
-vim.cmd [[
+vim.cmd([[
   highlight default link CmpItemAbbr CmpItemAbbrDefault
   highlight default link CmpItemAbbrDeprecated CmpItemAbbrDeprecatedDefault
   highlight default link CmpItemAbbrMatch CmpItemAbbrMatchDefault
   highlight default link CmpItemAbbrMatchFuzzy CmpItemAbbrMatchFuzzyDefault
   highlight default link CmpItemKind CmpItemKindDefault
   highlight default link CmpItemMenu CmpItemMenuDefault
-]]
```
hrsh7th, line 102: Thank you for your contribution!

But I don't know the reason why this code block doesn't touch on this PR. Could you explain it?

mehalter: That would be because I completely forgot about it, my bad! I will update this PR soon! 😂

mehalter: @hrsh7th I have updated my commit to resolve this issue

mehalter:
I am going to resolve this merge conflict and update my commit

hrsh7th:
Thank you!

Issue #936, opened 2022-04-25
Indications in the `documentation` deprecation notice cause an error


wincent:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/15f08a8faa22d52480cdcb9ef9ca698120f04363/doc/cmp.txt#L616) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/hrsh7th/nvim-cmp/issues) and there are no open or closed issues with the same problem.

### Neovim Version

NVIM v0.7.0

### Minimal reproducible full config

```
cmp.setup {
  documentation = { border = 'single' },
}
```

### Description

If you upgrade to a recent version of nvim-cmp containing old config like the above, it [prints](https://github.com/hrsh7th/nvim-cmp/blob/433af3dffce64cbd3f99bdac9734768a6cc41951/lua/cmp/config.lua#L194-L205) a deprecation notice of the form:

    [nvim-cmp] documentation is deprecated.
    [nvim-cmp] Please use window.documentation = "native" instead.

If you follow this advice literally, ie. with:

```
cmp.setup {
  window = {
    documentation = 'native',
  },
}
```

You'll get errors like these due to `documentation.max_width` being `nil`:

```
Error executing vim.schedule lua callback: ...nvim/pack/bundle/opt/nvim-cmp/lua/cmp/view/docs_view.lua:38: bad argument
     #1 to 'min' (number expected, got nil)
    stack traceback:
            [C]: in function 'min'
            ...nvim/pack/bundle/opt/nvim-cmp/lua/cmp/view/docs_view.lua:38: in function 'open'
            ...t/.config/nvim/pack/bundle/opt/nvim-cmp/lua/cmp/view.lua:229: in function 'callback'
            ...ig/nvim/pack/bundle/opt/nvim-cmp/lua/cmp/utils/async.lua:95: in function ''
            vim/_editor.lua: in function <vim/_editor.lua:0>
```

### Steps to reproduce

As above.

### Expected behavior

Deprecation notice should include a config suggestion that will work without errors.

### Actual behavior

Described above.

### Additional context

The workaround is to provide a well-formed and complete config object; eg:

```
cmp.setup {
  window = {
    documentation = cmp.config.window.bordered(),
  },
}
```

jc00ke:
Thanks for the workaround! :pray:

wookayin:
Related: #920 (actually the same bug)

[...]

PR #946, opened 2022-04-28
fix: don't show non-working configuration in deprecation suggestion

wincent:
As reported here: https://github.com/hrsh7th/nvim-cmp/issues/936


hrsh7th:
Thank you!

PR #952, opened 2022-05-01
Help page changes.

Uefair:
As a start, will say I adore that plugin! But disturbing thing was that docs aren't comprehensive or correct sometimes(no offense), that's why i attempted to fight against it and rewritten huge part of the doc.

### Aims of the PR

- Be short and expressive
- Follow grammar rules
- Maintain default structure of help file


hrsh7th:
Thank you!!!

PR #955, opened 2022-05-02
fix: cmdline fallback mapping for <S-Tab>

### LINKED ISSUE #953: Incorrect cmdline default mapping for `<S-Tab>`
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/15f08a8faa22d52480cdcb9ef9ca698120f04363/doc/cmp.txt#L616) and it didn't resolve my problem.

### Issues

### Neovim Version

0.7

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
  },

  sources = {
    { name = "nvim_lsp" },
    { name = "buffer" },
  },
}

cmp.setup.cmdline(':', {
    { name = 'path' }
  }, {
    { name = 'cmdline' }
  })
})

EOF

lua << EOF

  capabilities = capabilities,
}
EOF

```

### Description

Pressing `<S-Tab>` performs the same action as `<Tab>`, even though separate functions are called. This is due to https://github.com/hrsh7th/nvim-cmp/blob/baa8646c248486fdd46e8a087ae5623165901b9f/lua/cmp/config/mapping.lua#L65

### Steps to reproduce

`:P<tab>` brings up the default wildmenu (not configured with cmp). Pressing tab again will cycle to next entries.

### Expected behavior

Pressing shift tab should move you to previous entries.

### Actual behavior

Pressing shift tab cycles you to next entries (same as Tab)

### Additional context

Also, just noticed this from the README, but is there a difference between the `cmp.config.sources` call in https://github.com/hrsh7th/nvim-cmp/blob/baa8646c248486fdd46e8a087ae5623165901b9f/README.md#L112 and https://github.com/hrsh7th/nvim-cmp/blob/baa8646c248486fdd46e8a087ae5623165901b9f/README.md#L120?

axieax:
Closes #953

### DIFF
--- a/lua/cmp/config/mapping.lua
+++ b/lua/cmp/config/mapping.lua
@@ -62,7 +62,7 @@ mapping.preset.cmdline = function(override)
         if cmp.visible() then
           cmp.select_prev_item()
         else
-          feedkeys.call(keymap.t('<C-z>'), 'n')
+          feedkeys.call(keymap.t('<S-Tab>'), 'n')
         end
       end,
     },


hrsh7th:
In my environment, this fix breaks `<S-Tab>` mapping.

I think the following keys is valid... WDYT?

```lua
    ['<Tab>'] = {
      c = function()
        local cmp = require('cmp')
        if cmp.visible() then
          cmp.select_next_item()
        elseif vim.fn.pumvisible() == 1 then
          feedkeys.call(keymap.t('<Tab>'), 'n')
        else
          feedkeys.call(keymap.t('<C-z><C-n>'), 'n')
        end
      end,
    },
    ['<S-Tab>'] = {
      c = function()
        local cmp = require('cmp')
        if cmp.visible() then
          cmp.select_prev_item()
        elseif vim.fn.pumvisible() == 1 then
          feedkeys.call(keymap.t('<S-Tab>'), 'n')
        else
          feedkeys.call(keymap.t('<C-z><C-p>'), 'n')
        end
      end,
    },
```

axieax:
Hmm in my environment `<Tab>` skips an entry (moves twice instead of once), and `<S-Tab>` still doesn't work for some reason

https://user-images.githubusercontent.com/62098008/166261764-6ce5fa86-d135-4495-8942-480e10e36146.mp4

.

lua/cmp/config/mapping.lua
```diff
@@ -62,7 +62,7 @@ mapping.preset.cmdline = function(override)
         if cmp.visible() then
           cmp.select_prev_item()
         else
-          feedkeys.call(keymap.t('<C-z>'), 'n')
+          feedkeys.call(keymap.t('<S-Tab>'), 'n')
```
eyalk11, line 65: feedkeys is problematic as sometimes people change the default. `normal!` is the way.

axieax: @eyalk11 do you have an example for how would that be used here?

PR #956, opened 2022-05-02
dev

hrsh7th:


hrsh7th:
Thank you for reporting.
I would appreciate it if you could create an issue from the next time.

hermitmaster:
Sorry, I meant to create an issue.

wookayin:
So starting from this commit 5054c14, nvim-cmp requires lua v0.7.0? Why don't you please announce in #231 or make some warning messages?

smjonas:
Seems like it was announced in #231 (the edit from 9 days ago used the `status: applied`)

wookayin:
Also, I think when this branch was merged, it should have NOT squashed all the commits (bda03c9, a292926, b0896bf) as each of these has individual scope of changeset that is worth retaining the history. Such long-spanning branch can be simply fast forwarded or merged by rebasing multiple commits. IMO I believe those workflow will make the commit history much more readable, searchable, and easy to troubleshoot with when something goes wrong.

Issue #961, opened 2022-05-03
Invalid color settings passed to nvim_set_hl


IsaacClank:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/15f08a8faa22d52480cdcb9ef9ca698120f04363/doc/cmp.txt#L616) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/hrsh7th/nvim-cmp/issues) and there are no open or closed issues with the same problem.

### Neovim Version

v0.8.0-dev

### Minimal reproducible full config

```lua
local packer_repo = 'https://github.com/wbthomason/packer.nvim'
local packer_installation_path = vim.fn.stdpath('data') .. '/site/pack/packer/start/packer.nvim'

vim.fn.system({ 'git', 'clone', '--depth=1', packer_repo, packer_installation_path })
vim.cmd [[packloadall]]

require('packer').startup(function(use)
	use {
		'wbthomason/packer.nvim',
		{
			'neovim/nvim-lspconfig',
			requires = { 'williamboman/nvim-lsp-installer' }
		},
		{
			'hrsh7th/nvim-cmp',
			requires = {
				{ 'L3MON4D3/LuaSnip' },
				{ 'hrsh7th/cmp-buffer' },
				{ 'hrsh7th/cmp-cmdline' },
				{ 'hrsh7th/cmp-nvim-lsp' },
				{ 'hrsh7th/cmp-path' },
				{ 'hrsh7th/nvim-cmp' },
				{ 'saadparwaiz1/cmp_luasnip' },
			},
		},
		{ 'nvim-treesitter/nvim-treesitter' },
	}

	require('packer').sync()
	require('packer').compile()
end)

vim.api.nvim_create_autocmd('User', {
	pattern = 'PackerCompileDone',
	callback = function()
		local completion = require('cmp')
		local snippet = require('luasnip')

		local function has_words_before()
			local line, col = unpack(vim.api.nvim_win_get_cursor(0))
			return col ~= 0 and vim.api.nvim_buf_get_lines(0, line - 1, line, true)[1]:sub(col, col):match("%s") == nil
		end

		local function on_tab(fallback)
			if completion.visible() then
				completion.select_next_item()
			elseif snippet.expand_or_jumpable() then
				snippet.expand_or_jump()
			elseif has_words_before() then
				completion.complete()
			else
				fallback()
			end
		end

		completion.setup({
[...]

PR #962, opened 2022-05-04
fix: Invalid color settings passed to nvim_set_hl 

npmaile:
this is a fix to this issue:
https://github.com/hrsh7th/nvim-cmp/issues/961
Not sure if it's the correct fix, but from what we (the members of the @theprimeagen discord server @navigationhazard and @mvargasmoran) seem to think it's correct

### DIFF
--- a/lua/cmp/utils/highlight.lua
+++ b/lua/cmp/utils/highlight.lua
@@ -20,7 +20,7 @@ highlight.inherit = function(name, source, settings)
         v = v == 1
       end
       if v then
-        settings[key] = v == '' and 'NONE' or v
+        settings[key] = v == '' or 'NONE' or v
       end
     end
   end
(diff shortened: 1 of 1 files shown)


npmaile:
addresses https://github.com/hrsh7th/nvim-cmp/issues/961

lua/cmp/utils/highlight.lua
```diff
@@ -20,7 +20,7 @@ highlight.inherit = function(name, source, settings)
         v = v == 1
       end
       if v then
-        settings[key] = v == '' and 'NONE' or v
+        settings[key] = v == '' or 'NONE' or '0'
```
hrsh7th, line 23: `'NONE' or '0'` always return 'NONE'.

hrsh7th:
Thank you.
I can merge this for now. But I think it still not corrext to me.

hrsh7th:
Ah! Sorry. I'm getting the free wi-fi spot. I'll double check about this issue.

PR #965, opened 2022-05-04
add events: view_opened and view_closed

tzachar:
Add more events.

This allows the following behaviour (notice the up/down arrow pointing to the direction of the menu (top_down or bottom_up):

Peek 2022-05-04 15-49

Here is the code needed to implement it, assuming you are using the `custom` view, and setting `selection_order` to `near_cursor`:

```lua
local ns_id = vim.api.nvim_create_namespace('arrows')
local function remove_marks()
    local all = vim.api.nvim_buf_get_extmarks(0, ns_id, 0, -1, {})
    for _, mark in ipairs(all) do
      vim.api.nvim_buf_del_extmark(0, ns_id, mark[1])
    end
end

cmp.event:on('view_closed',
  function()
    remove_marks()
  end
)

cmp.event:on('view_opened',
  function(evt)
    if vim.api.nvim_get_mode().mode:sub(1, 1) == 'c' then
      return
    end
    local character
    local cursor = vim.api.nvim_win_get_cursor(0)
    local row = cursor[1] - 1
    local col = cursor[2]

    remove_marks()

    if evt.window.bottom_up then
      character = "↑"
    else
      character = "↓"
    end
    local opts = {
      id = 1,
      virt_text = {{character, "IncSearch"}},
      virt_text_pos = 'overlay',
      priority = 10005,
    }

    vim.api.nvim_buf_set_extmark(0, ns_id, row, col, opts)
  end
)


```

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -223,9 +223,9 @@ NOTE: `<Cmd>lua require('cmp').complete()<CR>` can be used to call these functio
   - `complete_done`: emit after current completion is done.
   - `confirm_done`: emit after confirmation is done.
   - `view_opened`: emit after opening a new completion window. Called with a table holding a key
-    named `window`, poining to the completion window implementation.
+    named `window`, pointing to the completion window implementation.
   - `view_closed`: emit after closing pum is closed. Called with a table holding a key
-    named `window`, poining to the completion window implementation.
+    named `window`, pointing to the completion window implementation.
 
 ==============================================================================
 Mapping                                                            *cmp-mapping*
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -222,10 +222,10 @@ NOTE: `<Cmd>lua require('cmp').complete()<CR>` can be used to call these functio
 
   - `complete_done`: emit after current completion is done.
   - `confirm_done`: emit after confirmation is done.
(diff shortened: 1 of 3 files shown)


wookayin:
I think one usual convention is to UpperCase for vim events. Like `CmpViewOpened`, etc.

tzachar:
> I think one usual convention is to UpperCase for vim events. Like `CmpViewOpened`, etc.

All the other events in cmp are lower_case.

hrsh7th:
I'm considering the `view` naming convention.
`entries` or `menu` is better, I think. (But we didn't use `menu` in other places...)

hrsh7th:
`docs` -> documentation window
`menu` -> completion menu window
`ghost_text` -> ghost text

If we choose this, I think we should refactor the internal codes as well.

tzachar:
I don't really have an opinion here... 
And what would you want to refactor?

hrsh7th:
OK. I understand what you don't have strong opinion here.

Could you rename the `view ` as `menu`? (e.g. `view_opened` -> `menu_opened`).

tzachar:
@hrsh7th 
Can we merge this?

hrsh7th:
LGTM.

PR #974, opened 2022-05-07
fix: configuration in README.md

Chaitanyabsprip:


hrsh7th:
I think the `.config` is not needed.

Chaitanyabsprip:
It didn't work without the .config so 🤷🏻‍♂️

Shougo:
I have tested it and it works without the `.config`.
Please check your version is the latest.

PR #991, opened 2022-05-15
Reuse scratch buffer for keymap normalization

tomtomjhj:
`keymap.normalize` is called many times when creating a buffer, so I think it makes sense to reuse the scratch buffer for it.


hrsh7th:
LGTM!

PR #1003, opened 2022-05-20
feat: Custom item kinds

### LINKED ISSUE #985: Add copilot icon
image

zbirenbaum:
Fixes #985 among others

This PR allows sources to return custom completion item kinds by returning a string instead of a int in the kind field. Currently, if you have a source which does not fit into the categories of a normal snippet or lsp kind there is no way to add an additional field. This fixes that. Here is an example in my copilot plugin:
```
source.format_completions = function(_, params, completions)
  return {
    label = label,
    kind = "Copilot",
    ...
  }
end
```

The benefits to this implementation (besides being 2 lines of code) are that highlight groups will be automatically added, for instance:
`vim.api.nvim_set_hl(0, CmpItemKindCopilot, '#A5BA93')`

Since cmp doesn't read config for formatting until immediately after determining type, this is by far the easiest way of making everything work in a non-invasive manner.

lspkind users will also be happy to know that I tested adding a custom group to it, and it works fine there as well:
```
require('lspkind').init({
  symbol_map = {
    Copilot = "[ﯙ]",
  },
})
```

I looked at a lot of different ways of doing this, but any method based on adding another entry and passing `kind = 26` for instance, resulted in hardcoding or overcomplication.

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -269,10 +269,13 @@ entry.get_vim_item = function(self, suggest_offset)
       end
     end
 
+    local cmp_opts = self:get_completion_item().cmp or {}
+
     local vim_item = {
       word = word,
       abbr = abbr,
-      kind = types.lsp.CompletionItemKind[self:get_kind()] or types.lsp.CompletionItemKind[1],
+      kind = cmp_opts.kind_text or types.lsp.CompletionItemKind[self:get_kind()] or types.lsp.CompletionItemKind[1],
+      kind_hl_group = cmp_opts.kind_hl_group,
       menu = menu,
       dup = self:get_completion_item().dup or 1,
     }
--- a/lua/cmp/types/lsp.lua
+++ b/lua/cmp/types/lsp.lua
@@ -170,6 +170,10 @@ lsp.CompletionItemKind = vim.tbl_add_reverse_lookup(lsp.CompletionItemKind)
 ---@field public detail string|nil
 ---@field public description string|nil
 
+---@class lsp.Cmp
+---@field public kind_text string
+---@field public kind_hl_group string
+
 ---@class lsp.CompletionItem
 ---@field public label string
 ---@field public labelDetails lsp.CompletionItemLabelDetails|nil
@@ -189,6 +193,7 @@ lsp.CompletionItemKind = vim.tbl_add_reverse_lookup(lsp.CompletionItemKind)
 ---@field public commitCharacters string[]|nil
 ---@field public command lsp.Command|nil
 ---@field public data any|nil
+---@field public cmp lsp.Cmp|nil
 ---
 ---TODO: Should send the issue for upstream?
 ---@field public word string|nil


zbirenbaum:
@hrsh7th Would it be possible for you to review this? Beyond the shown use case I need the functionality to write some custom comparators for ranking completions for just the copilot source.

hrsh7th:
The LSP specification defines [CompletionItemKind] (https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemKind).

I don't think it's a good idea for the source to return a non-compliant response.

For example, nvim-cmp already uses the `CompletionItemKind` as highlight definition.
The kind highlight definisions are also used from some of the colorschemes.

This feature is not fit current ecosystem, I think... (so I worried about merghing this)

max397574:
I think you could add other kinds
because this isn't just for lsp (as far as I understood)

hrsh7th:
I think we shouldn't add the hack but I can add the protocol extension point.

What do you about the following solution?

```lua
{
  label = 'if ... then`
  kind = lsp.CompletionItemKind.Text,
  cmp = {
    kind_text = 'Copilot'
  }
}
```

zbirenbaum:
> I think we shouldn't add the hack but I can add the protocol extension point.
> 
> 
> 
> What do you about the following solution?
> 
> 
> 
> ```lua
> 
> {
> 
>   label = 'if ... then`
> 
>   kind = lsp.CompletionItemKind.Text,
> 
>   cmp = {
> 
>     kind_text = 'Copilot'
> 
>   }
> 
> }
> 
> ```

How would it work to differentiate between Text responses for highlights or adding icons?

hrsh7th:
This is nothing to difference. We can do the same thing to achive.

```
require('lspkind').init({
  symbol_map = {
    Copilot = "[ﯙ]",
  },
})
```

Your suggested solution modifys the `kind` field type. It's defined as `number` type but this PR make to allow `string` type value.

My solution does not modify the LSP.CompletionItem definition but adds ability to customize `vim_item.kind` value. (In my knowlege, lspkind uses this value.

zbirenbaum:
I can modify the PR to do that instead, but there is still the issue of hl_groups. Is it alright with you if I also modify it to allow for:
```
{
  label = 'if ... then`
  kind = lsp.CompletionItemKind.Text,
  cmp = {
    kind_hl_group = "CmpItemKindCopilot",
    kind_text = 'Copilot'
  }
}
```

hrsh7th:
I think we can have the `kind_hl_group`.

zbirenbaum:
@hrsh7th I implemented it as you described, and it is working. vim_item.kind must be set to be the value of cmp.kind_text, rather than using view.kind.text for lspkind to work with this change. This is because the format function is called on the `vim_item` immediately after its creation in `entry.get_vim_item`

I added the fields and classes where necessary inside of the type declarations, but you may want to double check that I did it exactly how you would prefer.

zbirenbaum:
@hrsh7th Just wanted to check in on this, it's a bit outdated now but I can rebase if you like. Did you have a chance to look over the implementation and see if it matches what you were looking for?

lua/cmp/types/vim.lua
```diff
@@ -10,6 +10,7 @@
 ---@field public abbr_hl_group string|nil
 ---@field public kind_hl_group string|nil
 ---@field public menu_hl_group string|nil
+---@field public cmp table|nil
```
hrsh7th, line 13: I think it isn't needed.

hrsh7th:
LGTM. The last point is unnecessary type hints. Could you remove it?

_Sorry for the late review._

lua/cmp/types/vim.lua
zbirenbaum: Got it, removed that line.

hrsh7th:
LGTM!

PR #1006, opened 2022-05-22
docs: add minimal_init.lua

kylo252:
I saw that this is a common request, so I figured I can create a minimal template in Lua.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -20,17 +20,20 @@ body:
 
   - type: textarea
     attributes:
-      label: "Minimal reproducible full config"
+      label: "Steps to reproduce"
+      description: "Steps to reproduce using the minimal config provided below."
+      placeholder: |
+        1. `nvim -u minimal_init.lua`
+        2. ...
+    validations:
+      required: true
+
+  - type: textarea
+    attributes:
+      label: "Minimal config"
+      render: Lua
       description: |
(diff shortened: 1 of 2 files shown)


David-Else:
I can recommend https://github.com/savq/paq-nvim , very minimal.

kylo252:
> I'm not familiar with packer.nvim, but I've seen packer.nvim cause caching issues. Is there a cache problem with this minimum.lua?
 
I'm not sure, but I don't think there are generally any such issues

cc: @wbthomason, @akinsho

akinsho:
I'm not sure what the caching problems specifically relate to, but all I can provide is some general clarification on what people often perceive as caching.

1. Usage of native vim packages - packer (and paq) use native vim packages which were added in vim 8 which means any plugins installed into the `start` directory which will be the case for anything not conditionally loaded will automatically *be loaded by neovim on startup*, packer does not control this behaviour

2. The packer compiled file, this is only really relevant for conditionally loading plugins using autocommands or nvim commands and is not really relevant to start plugins. This file is read on startup and registers autocommands etc. to conditionally load plugins

I believe the caching issues are largely due to confusion around the behaviour of vim packages which differs from something like `vim-plug` which leads users to believe that packer is doing something rather than reverse which is that vim-plug works differently to native behaviour 🤷🏿 

From what I can see there is no conditional loading here, so I don't know any reason why there would be any caching related issues

Shougo:
I think the minimal init.lua or init.vim should not be depend on any of plugin mangers.

PR #1011, opened 2022-05-25
fix: only scroll docs if docs view is visible

williamboman:
I ran into some issues with indenting/unindenting code via `<C-d>/<C-t>` when the completion menu was open but no docs window visible.

I was also a bit surprised to see that when indenting/unindenting code it also opens the completion menu, which feels like a potential separate issue?


hrsh7th:
That's make sense. Thank you!

PR #1012, opened 2022-05-25
feat: add ability to send docs popup to a preview window

williamboman:
My searching might have been really bad, but I couldn't find any discussions around a feature like this. Apologies for jumping straight to implementation but the source code was very well written and easy to get into so I couldn't help myself :).

Demo: https://asciinema.org/a/sJWBBt3lQKs6NFyf3HdlqNCsU

---

So the idea behind this is to give users the ability to interact with the docs contents in more lasting contexts than just a popup window that closes itself, by providing the ability to send it to a new preview window (or overwrite the contents of any existing preview window). What do you think?

### DIFF
--- a/lua/cmp/view/docs_view.lua
+++ b/lua/cmp/view/docs_view.lua
@@ -1,5 +1,6 @@
 local window = require('cmp.utils.window')
 local config = require('cmp.config')
+local buffer = require('cmp.utils.buffer')
 
 ---@class cmp.DocsView
 ---@field public window cmp.Window
@@ -157,6 +158,8 @@ docs_view.send_to_preview_window = function(self)
   vim.cmd([[stopinsert]])
   local pwin = get_preview_window()
   vim.api.nvim_win_set_buf(pwin, self.window:get_buffer())
+  buffer.cache[self.window.name] = nil -- release buffer cache as it's now managed manually by the user
+  self.window:close()
 end
 
 return docs_view
--- a/lua/cmp/view/docs_view.lua
+++ b/lua/cmp/view/docs_view.lua
@@ -16,6 +16,12 @@ docs_view.new = function()
   self.window:option('linebreak', true)
   self.window:option('scrolloff', 0)
   self.window:option('wrap', true)
+  self.window:buffer_option('modifiable', false)
+  self.window:buffer_option('undolevels', -1)
+  self.window:buffer_option('buflisted', false)
+  self.window:buffer_option('buftype', 'nofile')
+  self.window:buffer_option('bufhidden', 'wipe')
+  self.window:buffer_option('swapfile', false)
   return self
 end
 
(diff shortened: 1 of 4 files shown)


lua/cmp/view/docs_view.lua
```diff
@@ -16,6 +16,12 @@ docs_view.new = function()
   self.window:option('linebreak', true)
   self.window:option('scrolloff', 0)
   self.window:option('wrap', true)
+  self.window:buffer_option('modifiable', false)
+  self.window:buffer_option('undolevels', -1)
+  self.window:buffer_option('buflisted', false)
+  self.window:buffer_option('buftype', 'nofile')
+  self.window:buffer_option('bufhidden', 'wipe')
+  self.window:buffer_option('swapfile', false)
```
williamboman, line 24: I figured it'd make sense to set these buffer options for the docs window.

lua/cmp/view/docs_view.lua
```diff
@@ -47,12 +53,14 @@ docs_view.open = function(self, e, view)
...
+      vim.lsp.util.stylize_markdown(0, documents, {
+        max_width = max_width,
+        max_height = documentation.max_height,
+      })
+      vim.api.nvim_buf_set_option(0, 'modifiable', false)
     end)
-    vim.lsp.util.stylize_markdown(self.window:get_buffer(), documents, {
-      max_width = max_width,
-      max_height = documentation.max_height,
-    })
```
williamboman, line 55: This change was primarily needed because we need to toggle the `modifiable` buffer option while settings its contents. Since there's already a call to `nvim_buf_call` I figured all buffer modifications could be moved inside that callback

hrsh7th:
I think this feature shouldn't included in to the nvim-cmp... maybe.

I think it can be more general plugins that enables same thing on `textDocument/hover`. WDYT?

williamboman:
> I think it can be more general plugins that enables same thing on `textDocument/hover`. WDYT?

I'm not sure `textDocument/hover` and `textDocument/completion` fully overlap each other in terms of the documentation they provide. There might be `CompletionItem`s that don't really have a `textDocument/hover` counter-part (snippets for example).

I guess it'd make sense for a generic way of sending any kind of "documentation popup window currently visible" to the preview window, these buffers/windows would then have to be identifiable/tagged somehow, no? Or did you have some other idea?

williamboman:
I ran into another compelling use case, opening the docs window of the `cmp-git` extension: https://asciinema.org/a/WI2EeANXkJ1tQoYPd5PqKU6kO.

max397574:
imo there is no reason not to include this in cmp
it's something really useful and really small

max397574:
@williamboman does this already work?
(is it a draft because it's not finished or because it's not done yet?)

williamboman:
This works. I've kept it in draft because there hadn't been any prior discussions whether something like this should even make it into cmp core. I'll leave that decision to @hrsh7th!

hrsh7th:
If I address this issue, I would develop a separate plugin like `nvim-float2preview`.
I think it should also apply to LSP hover and cmp docs window or any other floating windows.

But I don't have enough time to develop such plugin at the moment...

So I'll review this PR for merging.
Anyway, Greatly thank you for @williamboman

hrsh7th:
@max397574 
I think bumping "a feature should be in the core" is basically not recommended.

For example, I'm using the LSP diagnostics with a floating-window. I've also wanted to send it to a preview.

Don't you think that sending an nvim-cmp window to a preview and sending a diagnostics window to a preview should ideally be provided by a single plug-in in a single workflow?

_Google translated_

williamboman:
Makes sense! I'll look into extracting my workflow into a dedicated plugin

Sleepful:
@williamboman hi, did you find a way to implement this behavior with `nvim-cmp`? Yearning for the same thing here.

williamboman:
> @williamboman hi, did you find a way to implement this behavior with `nvim-cmp`? Yearning for the same thing here.

No, I keep running this feature branch and occasionally merge with upstream :D

Shougo:
If the branch is separated extension, it is easy to use.
I don't know it is possible though.

Sleepful:
hmm! A plugin would be cool! 

I saw that you thought about it for two seconds in the `hover.nvim` issue tracker 😄 
So I installed the `hover.nvim`, I guess just for the borders on the hover windows, looks nice.

I suppose right now the same result can be done in an extra step or two:

Confirm the selected suggestion, press `K` to display docs in hover window. Seems like that is more or less the same thing? I guess if you end up deciding that you wanted another auto complete you also have to backspace the word and try again.

I am curious if you think you are getting more mileage from the feature than I just described @williamboman ?

PR #1026, opened 2022-06-06
make debounce and throttle time configurable

MurdeRM3L0DY:
supersedes #639.

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -366,6 +366,16 @@ enabled~
   `boolean | fun(): boolean`
   Toggles the plugin on and off.
 
+                                                *cmp-config.performance.debounce*
+performance.debounce~
+  `number`
+  Sets debounce time
+
+                                                *cmp-config.performance.throttle*
+performance.throttle~
+  `number`
+  Sets throttle time
+
                                                           *cmp-config.preselect*
 preselect~
   `cmp.PreselectMode`
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -370,11 +370,14 @@ enabled~
 performance.debounce~
   `number`
   Sets debounce time
+  This is the interval used to group up completions from different sources
+  for filtering and displaying.
 
                                                 *cmp-config.performance.throttle*
 performance.throttle~
   `number`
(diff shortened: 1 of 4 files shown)


doc/cmp.txt
```diff
@@ -366,6 +366,16 @@ enabled~
   `boolean | fun(): boolean`
   Toggles the plugin on and off.
 
+                                                *cmp-config.performance.debounce*
+performance.debounce~
+  `number`
+  Sets debounce time
```
hrsh7th, line 372: I think we should add more obvious description for this.

doc/cmp.txt
```diff
@@ -366,6 +366,16 @@ enabled~
...
+
+                                                *cmp-config.performance.throttle*
+performance.throttle~
+  `number`
+  Sets throttle time
```
hrsh7th, line 377: ditto

This value is used for throttling `filtering` and `displaying` results.
For example, nvim-cmp shouldn't filter results 7 times if the user press `console` very fast.
If the user specify `throttle_time` as 120(ms), nvim-cmp will filter results each 70ms.

doc/cmp.txt
hrsh7th: Hm... the `debounce` is hard to explain.

If the user press `console` and it's invoking 10 number of sources, the nvim-cmp will retrieve completion results 10 times.

Typically, the each responses would come back from sources in small milliseconds.
I think it should be merged as one filtering process. The debounce_times is used for this.

MurdeRM3L0DY: So debounce is basically how long cmp waits to retrieve completion results from sources before displaying.

MurdeRM3L0DY:
WDYT? @hrsh7th

hrsh7th:
I can merge this after docmentation updated.

hrsh7th:
LGTM. Thank you~!

MurdeRM3L0DY:
> LGTM. Thank you~!

Happy to contribute. Thank YOU for this amazing plugin!

hrsh7th:
Yes. we can close that. I'll do.

PR #1037, opened 2022-06-16
fix: nvim_create_user_command to nvim_add_user_command

vctrblck:
`vim.api.nvim_add_user_command` seems to be the working definition according to NeoVim v0.7.0 docs. 

`vim.api.nvim_create_user_command` throws an E5113 error for referencing a nil value.

### DIFF
--- a/plugin/cmp.lua
+++ b/plugin/cmp.lua
@@ -52,7 +52,7 @@ if vim.on_key then
   end, vim.api.nvim_create_namespace('cmp.plugin'))
 end
 
-vim.api.nvim_create_user_command('CmpStatus', function()
+vim.api.nvim_add_user_command('CmpStatus', function()
   require('cmp').status()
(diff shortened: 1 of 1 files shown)


max397574:
this will break things for nightly users

max397574:
nvim_create_user_command it what's used on latest stable
so you should just update to it

hrsh7th:
@vctrblck Could you add the feature detection for this?

```lua
-- TODO: remove after nvim v0.8.0 is released.
(vim.api.nvim_create_user_command or vim.api.nvim_add_user_command)('CmpStatus', function()
  require('cmp').status()
end, { desc = 'Check status of cmp sources' })
```

max397574:
as I said above
nvim_create_user_command is the only right one
for 0.7 and 0.8
no checks needed

PR #1045, opened 2022-06-23
Fix type comment for cmp.ContextReason

tristan957:
Side note, where is this documented? I see sumneko warning me about not providing an argument to cmp.mapping.complete(), but I don't know what the options object even does.


hrsh7th:
very sorry. This is a bit of an internal value and has not been documented yet.

I think it should be documented because it is used by some APIs, but I haven't decided on a policy for documenting Lua type definitions.

_google translated_

Issue #1041, opened 2022-06-17
Filter items by kind?


joshfrench:
Is it possible to filter out items by kind? For example, I might want to see `Text` items that come from `cmp-buffer` but **not** any `Text` items from `cmp-nvim-lsp`.

Thanks for a great plugin!

hrsh7th:
The entry_filter option can be used now.

Issue #1036, opened 2022-06-16
Ignore certain words/patterns


mway:
Thanks for the great plugins!

One issue that's been frustrating a lot lately is that there are unwanted completions. For a trivial example, in Go, if I'm writing a function like the following:

```go
func doSomething() error {
  return nil
}
```

after I've typed `return ni`, suggestions begin popping up (I use `keyword_length=2` and `<CR>` to confirm, fyi), but the very last one is `nil`. Instead of `nil` being at the top (which I'd expect because it's the closest match), the first suggestion might be `nilerr`, which is a package that's (probably) a very transitive dependency that I don't use in my project, and so it'll be selected when I hit return (I type fast), and gopls will add an import for it as well, which I then need to undo.

Anyway, I know I can use something like:

```lua
cmp.mapping.confirm({ select = false })
```

so that I'd need to select it explicitly, but what I'd really like is just to ignore `nilerr` (and various others) completely as an option. I considered adding a function handler for the `<CR>` bind, but they'd still show in the completion list, and I'd have to keep typing or navigate through them to get what I actually wanted in the first place.

Is this something that's already possible? If not, would you consider adding support for it?

Happy to provide any other info you'd like!

hrsh7th:
The entry_filter option can be used now.

Issue #806, opened 2022-02-18
Is it possible to have context-aware sources?


norseghost:
With the context support that was added recently, it is now possible to disable completion in say, comments. Awesome!

I want to have some sources in comments; and some outside. I tried creating a function that checked current context and set sources appropriately; and thought it worked - but this didn’t update on context change while editing:

```lua

-- cmp setup lua file
local context = require("cmp.config.context")
local set_sources = function()
    local s = {}
    if context.in_syntax_group('Comment') or context.in_treesitter_capture('comment') then
        s = {
            -- comment sources
        }
    else
        s = {
            -- other sources
        }
    end
    return s
end

cmp.setup({
    -- other setup
    sources = set_sources(),
)}
```

ideally; I’d like to pass a function directly to cmp.setup.sources. Could I hack around this with an InsertEnter autocommand and cmp.setup.buffer maybe?

st3fan:
This would be really nice. I generally do not care about most sources in comments, but some may be relevant.

bennypowers:
x-post from https://github.com/hrsh7th/cmp-nvim-lsp/issues/20#issuecomment-1175254858

Take this javascript source:
```js
import { html, render } from 'lit-html';

render(html`
  <p>I'm a little teapot</p>
  
`, document.body);
```

With the cursor on the line below the `<p>` tag, I'd like to receive emmet completions from [emmet_ls](https://github.com/aca/emmet-ls), but when the cursor is on the line beneath the `import` statement, I certainly would not like to receive emmet completions.

I would like the ability to filter lsp snippets by treesitter context.

Could `SourceConfig` also take an `enabled` predicate?

kkharji:
> x-post from [hrsh7th/cmp-nvim-lsp#20 (comment)](https://github.com/hrsh7th/cmp-nvim-lsp/issues/20#issuecomment-1175254858)
> 
> Take this javascript source:
> 
> ```js
> import { html, render } from 'lit-html';
> 
> render(html`
>   <p>I'm a little teapot</p>
>   
> `, document.body);
> ```
> 
> With the cursor on the line below the `<p>` tag, I'd like to receive emmet completions from [emmet_ls](https://github.com/aca/emmet-ls), but when the cursor is on the line beneath the `import` statement, I certainly would not like to receive emmet completions.
> 
> I would like the ability to filter lsp snippets by treesitter context.
> 
> Could `SourceConfig` also take an `enabled` predicate?

with #1067 you can do 

[...]

PR #1067, opened 2022-07-07
feat: add `filter` option to sources

jonatan-branting:
This allows the user to specify a `filter` function for each source.

Solves:
https://github.com/hrsh7th/nvim-cmp/issues/1036
https://github.com/hrsh7th/nvim-cmp/issues/1041
https://github.com/hrsh7th/nvim-cmp/issues/806

@hrsh7th do you see any problems with this implementation or approach in general?

An example (Updated):
```lua
-- don't show entries with kind "Text" from the "nvim_lsp" source
sources = {
{
  name = "nvim_lsp",
  entry_filter = function(entry, ctx)
    local kind = types.lsp.CompletionItemKind[entry:get_kind()]

    if kind == "Text" then return false end
    return true
  },
}
```

By utilizing the `ctx` parameter, the user can also ignore certain
entries in certain contexts.

Notes:
- <s>The name `filter` might not be entirely clear; what does returning `true` or `false` actually result in? Ideas?</s> Update: Changed name to `entry_filter`, and swapped true/false, making it function more like traditional filter functions.
- Documentation has to be added.
- <s>Specs have to be added.</s> There doesn't seem to be a precedence for testing other configuration options, not sure it's entirely necessary either.

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -546,6 +546,38 @@ sources[n].group_index~
       })
     }
 <
+
+                                             *cmp-config.sources[n].entry_filter*
+sources[n].entry_filter~
+  `function`
+  A source-specific entry filter, with the following function signature:
+>
+  --- @param entry cmp.Entry
+  --- @param ctx cmp.Context
+  --- @return boolean
+  function(entry, ctx) end
+<
+
+  Returning `true` will keep the entry, while returning `false` will remove it.
+
+  This can be used to hide certain entries from a given source. For instance, you
+  could hide all entries with kind `Text` from the `nvim_lsp` filter using the
+  following source definition:
+>
+  {
+    name = 'nvim_lsp',
+    entry_filter = function(entry, ctx)
+      local kind = require('cmp.types').lsp.CompletionItemKind[entry:get_kind()]
+
+      if kind == 'Text' then return false end
+
+      return true
+    end
+  }
+<
+  Using the `ctx` parameter, you can further customize the behaviour of the
+  source.
+
                                                                *cmp-config.view*
 view~
   `{ entries: cmp.EntriesConfig|string }`
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -552,10 +552,7 @@ sources[n].entry_filter~
   `function`
   A source-specific entry filter, with the following function signature:
 >
-  --- @param entry cmp.Entry
-  --- @param ctx cmp.Context
-  --- @return boolean
-  function(entry, ctx) end
+  function(entry: cmp.Entry, ctx: cmp.Context): boolean
 <
 
   Returning `true` will keep the entry, while returning `false` will remove it.
@@ -567,9 +564,7 @@ sources[n].entry_filter~
   {
     name = 'nvim_lsp',
     entry_filter = function(entry, ctx)
-      local kind = require('cmp.types').lsp.CompletionItemKind[entry:get_kind()]
-
-      return kind ~= 'Text'
+      return require('cmp.types').lsp.CompletionItemKind[entry:get_kind()] ~= 'Text'
     end
   }
 <
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -569,9 +569,7 @@ sources[n].entry_filter~
     entry_filter = function(entry, ctx)
       local kind = require('cmp.types').lsp.CompletionItemKind[entry:get_kind()]
 
-      if kind == 'Text' then return false end
-
-      return true
+      return kind ~= 'Text'
     end
   }
 <
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -101,6 +101,8 @@ source.get_entries = function(self, ctx)
     return self.entries
   end)()
 
+  local entry_filter = self:get_entry_filter()
+
   local inputs = {}
   local entries = {}
   for _, e in ipairs(target_entries) do
@@ -115,7 +117,10 @@ source.get_entries = function(self, ctx)
     if e.score >= 1 then
       e.matches = match.matches
       e.exact = e:get_filter_text() == inputs[o] or e:get_word() == inputs[o]
-      table.insert(entries, e)
+
+      if entry_filter(e, ctx) then
+        table.insert(entries, e)
+      end
     end
   end
   self.cache:set({ 'get_entries', self.revision, ctx.cursor_before_line }, entries)
@@ -230,6 +235,16 @@ source.get_keyword_length = function(self)
     return c.keyword_length
   end
   return config.get().completion.keyword_length or 1
+end
+
+---Get filter
+--@return function
+source.get_entry_filter = function(self)
+  local c = self:get_source_config()
+  if c.entry_filter then
+    return c.entry_filter
+  end
+  return function(_, _) return true end
 end
 
 ---Invoke completion
--- a/lua/cmp/types/cmp.lua
+++ b/lua/cmp/types/cmp.lua
@@ -152,7 +152,7 @@ cmp.ItemField = {
 ---@field public keyword_length integer|nil
 ---@field public max_item_count integer|nil
 ---@field public group_index integer|nil
----@field public entry_filter function|nil
+---@field public entry_filter nil|function(entry: cmp.Entry, ctx: cmp.Context): boolean
 
 ---@class cmp.ViewConfig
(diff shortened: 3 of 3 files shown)


Weissle:
As a user, I think this feature is useful. I have luasnip and some LSP servers also provide snippets. It is annoying that LSP's snippets are always standing before the luasnip's but the quality of them is not so good. 
I know how to adjust their priority, but a filter function provides great scalability.

hrsh7th:
Yes. This feature is useful. I'm thinking about the API design...

Massolari:
Usually, in `filter` functions, when you return `true` that means you want the item, if you don't want the item, you return `false` .

lua/cmp/source.lua
```diff
@@ -115,7 +115,11 @@ source.get_entries = function(self, ctx)
     if e.score >= 1 then
       e.matches = match.matches
       e.exact = e:get_filter_text() == inputs[o] or e:get_word() == inputs[o]
-      table.insert(entries, e)
+
+      local filter = self:get_source_config().filter
+      if not filter or not filter(e, ctx) then
```
bennypowers, line 120:
```suggestion
      if type(filter) ~= 'function' or filter(e, ctx) then
```

if no filter was defined, insert
otherwise, if the filter was defined and it passes for the entry/context pair, insert

This could also be done by defaulting the filter to `() -> true`

gotgenes: >This could also be done by defaulting the filter to `() -> true`

+1 to this approach! I find setting a default filter of `() -> true` and then always applying the filter easier to understand. [Here's a talk on this approach](https://www.youtube.com/watch?v=AnZ0uTOerUI) by Michael Feathers.

jonatan-branting: I do like that approach better. I'd implement it, but I got the feeling that @hrsh7th cooking up something on his own.

Weissle: @hrsh7th seems not active recently.
But it would be better to have a discuss with he to avoid you are working on the same feature.

jonatan-branting: See: [03a7a91](https://github.com/hrsh7th/nvim-cmp/pull/1067/commits/03a7a91c7e1da6ce0147942a91b08fe3c52884d9)

jonatan-branting:
I pushed some updates according to feedback ([03a7a91](https://github.com/hrsh7th/nvim-cmp/pull/1067/commits/03a7a91c7e1da6ce0147942a91b08fe3c52884d9)). If anyone used my fork then this will have broken your configuration, sorry about that! @kkharji and @bennypowers

jonatan-branting:
@hrsh7th if this is good enough for you I'll add documentation and we can maybe get this merged? :)

lua/cmp/source.lua
```diff
@@ -115,7 +115,10 @@ source.get_entries = function(self, ctx)
     if e.score >= 1 then
       e.matches = match.matches
       e.exact = e:get_filter_text() == inputs[o] or e:get_word() == inputs[o]
-      table.insert(entries, e)
+
+      if self:get_entry_filter()(e, ctx) then
```
jonatan-branting, line 119: Move this out of the loop, fetching the filter for each iteration seems a tad unnecessary.

lua/cmp/source.lua
```diff
@@ -232,6 +235,16 @@ source.get_keyword_length = function(self)
...
 end
 
+---Get filter
+--@return function
+source.get_entry_filter = function(self)
+  local c = self:get_source_config()
+  if c.entry_filter then
+    return c.entry_filter
+  end
+  return function(_, _) return true end
```
bennypowers, line 245: nice

Weissle:
What about the document? @hrsh7th may not merge this PR without the document updated.

jonatan-branting:
> What about the document? @hrsh7th may not merge this PR without the document updated.

I'm holding off on that until I get the OK from @hrsh7th whether or not this is the solution he wants to go for or not.

hrsh7th:
I thought it would be more versatile to add an `override` option to the source and add the ability to override all source functions. As shown below.

```lua
{
  name = 'buffer',
  override = {
    complete = function(source_complete, params, callback)
      source_complete(params, function(response))
        -- modify complete response from source.
        callback(modify(response))
      end)
    end
  }
}
```

However, this API is too advanced.
Therefore, I can accept this feature.

Translated with www.DeepL.com/Translator (free version)

hrsh7th:
Please update the docs.

jonatan-branting:
> Please update the docs.

I'll most likely be able to get this done tomorrow :)

doc/cmp.txt
```diff
@@ -546,6 +546,38 @@ sources[n].group_index~
...
+  This can be used to hide certain entries from a given source. For instance, you
+  could hide all entries with kind `Text` from the `nvim_lsp` filter using the
+  following source definition:
+>
+  {
+    name = 'nvim_lsp',
+    entry_filter = function(entry, ctx)
+      local kind = require('cmp.types').lsp.CompletionItemKind[entry:get_kind()]
+
+      if kind == 'Text' then return false end
```
bennypowers, line 572:
```suggestion
      return kind ~= 'Text'
```

doc/cmp.txt
```diff
@@ -546,6 +546,38 @@ sources[n].group_index~
...
+
```
bennypowers, line 573:
```suggestion
```

doc/cmp.txt
```diff
@@ -546,6 +546,38 @@ sources[n].group_index~
...
+      return true
```
bennypowers, line 574:
```suggestion
```

jonatan-branting:
@hrsh7th I've added documentation, let me know if there is anything you want me to change!

doc/cmp.txt
jonatan-branting: That's some lightning-quick reviewing, thanks!

bennypowers (approved):
:heart:

bennypowers:
add this to line 155 in lua/types/cmp.lua

```lua
---@field public entry_filter function(entry: cmp.Entry, ctx: cmp.Context): boolean|nil
```

full block:
```lua
---@class cmp.SourceConfig
---@field public name string
---@field public option table|nil
---@field public priority integer|nil
---@field public trigger_characters string[]|nil
---@field public keyword_pattern string|nil
---@field public keyword_length integer|nil
---@field public max_item_count integer|nil
---@field public group_index integer|nil
---@field public entry_filter function(entry: cmp.Entry, ctx: cmp.Context): boolean|nil
```

doc/cmp.txt
```diff
@@ -546,6 +546,36 @@ sources[n].group_index~
...
+  This can be used to hide certain entries from a given source. For instance, you
+  could hide all entries with kind `Text` from the `nvim_lsp` filter using the
+  following source definition:
+>
+  {
+    name = 'nvim_lsp',
+    entry_filter = function(entry, ctx)
+      local kind = require('cmp.types').lsp.CompletionItemKind[entry:get_kind()]
+
+      return kind ~= 'Text'
```
hrsh7th, line 572: It's bit redundunt. Could you change as this?
```
return require('cmp').lsp.CompletionItemKind.Text ~= entry:get_kind()
```

lua/cmp/types/cmp.lua
```diff
@@ -152,6 +152,7 @@ cmp.ItemField = {
 ---@field public keyword_length integer|nil
 ---@field public max_item_count integer|nil
 ---@field public group_index integer|nil
+---@field public entry_filter function|nil
```
hrsh7th, line 155: You should add type annotation here.

jonatan-branting: See: [a7d6dd2](https://github.com/hrsh7th/nvim-cmp/pull/1067/commits/a7d6dd29c6ec3acd7bee514c4291aa5a062648f8)

doc/cmp.txt
jonatan-branting: See: [a7d6dd2](https://github.com/hrsh7th/nvim-cmp/pull/1067/commits/a7d6dd29c6ec3acd7bee514c4291aa5a062648f8)

Weissle:
Could you take a look? This featue have been backlogged for a month. @hrsh7th

Weissle (approved):

hrsh7th:
LGTM. Thank you.

PR #1075, opened 2022-07-14
Fix native_entries_view.info width and col calculations

gravndal:
The dictionary returned by `pum_getpos()` uses `scrollbar` not
`scrollable` to indicate whether the scrollbar is visible or not.

In addition, the padding on the left side of the popup menu isn't
included in the output of `pum_getpos()`, meaning that both `col` and
`width` (as understood by nvim-cmp) are off by one when the cursor isn't
at column `0`.

These two issues were causing the documentation window and popup menu to
overlap.


hrsh7th:
Thank you!

PR #1085, opened 2022-07-22
fix entry appearing for identical text

zbirenbaum:
Fixes zbirenbaum/copilot-cmp#15

The formatting required for copilot text completions causes cmp to sometimes have an entry remain in the completion menu after it is accepted leading to some sizable problems as can be seen in the above issue.

Debugging matcher shows the input and word parameters are identical, so I believe the core of this issue is actually a bug elsewhere in the source with label being prioritized over filterText. Tracking down where that might be occurring if it even is, is quite difficult, so a simple equivalency check alongside the existing length check suffices.

### DIFF
--- a/lua/cmp/matcher.lua
+++ b/lua/cmp/matcher.lua
@@ -88,8 +88,8 @@ matcher.match = function(input, word, option)
     return matcher.PREFIX_FACTOR + matcher.NOT_FUZZY_FACTOR, {}
   end
 
-  -- Ignore if input is long than word
-  if #input > #word then
+  -- Ignore if input is identical to or longer than word
+  if input == word or #input > #word then
     return 0, {}
   end
 


hrsh7th:
I think this is huge breaking change and don't think it's good solution.


For example, consider the case where the user types `useState` and then confirm it to trigger auto-import.

This PR doesn't allow this workflow because `useState` candidates will disappear.

zbirenbaum:
> I think this is huge breaking change and don't think it's good solution.
> 
> 
> 
> 
> 
> For example, consider the case where the user types `useState` and then confirm it to trigger auto-import.
> 
> 
> 
> This PR doesn't allow this workflow because `useState` candidates will disappear.
> 
> 

This was my concern since I wasn't sure what the side effects would be, and I probably should have labeled this as a draft. Do you have any input on what a better way to go about this would be?

hrsh7th:
If you want to "do not match when you type the word to the end", it seems to work if you put the word with one letter removed in filterText.

_google translated_

hrsh7th:
I tried it with `cmp-buffer` source. It works as you expected.

```
              filterText = string.sub(word, 1, #word - 1),
```

zbirenbaum:
> If you want to "do not match when you type the word to the end", it seems to work if you put the word with one letter removed in filterText.
> 
> _google translated_

this has been my temp fix, if that's your official recommendation I will just keep it there and close this

hrsh7th:
It is a bit annoying, but normal completion systems do not hide exact match candidates.

Therefore, we think it makes poor sense to include the new feature on the nvim-cmp side.

I think the only way to deal with this is to use filterText or to use isIncomplete to control it completely on the source side.

zbirenbaum:
> It is a bit annoying, but normal completion systems do not hide exact match candidates.
> 
> Therefore, we think it makes poor sense to include the new feature on the nvim-cmp side.
> 
> I think the only way to deal with this is to use filterText or to use isIncomplete to control it completely on the source side.
> 
> _deepl translated_

I will go ahead and close this in that case, but could you clarify what isIncomplete actually does? I can't seem to find any documentation for it and it would be helpful to know exactly what behavior it causes.

hrsh7th:
The isIncomplete is defined in LSP spec.

If you returned the `{ items: ..., isIncomplete: true }` via` source.complete`, the nvim-cmp will call `source.complete` every keystroke.

Typically, the completion engine consider `keyword_pattern` to invoke `source.completion`. The `isIncomplete` flag is to disable this.

Let's consider the case where the user types characters from `table.|` to `table.insert|`.

1. `table.i|`
    - nvim-cmp will invoke `source.complete`
2. `table.in|`
    - nvim-cmp does not invoke `source.complete` because `keyword_pattern` is `\w` typically.

zbirenbaum:
> The isIncomplete is defined in LSP spec.
> 
> If you returned the `{ items: ..., isIncomplete: true }` via` source.complete`, the nvim-cmp will call `source.complete` every keystroke.
> 
> Typically, the completion engine consider `keyword_pattern` to invoke `source.completion`. The `isIncomplete` flag is to disable this.
> 
> Let's consider the case where the user types characters from `table.|` to `table.insert|`.
> 
> 1. `table.i|`
>    
>    * nvim-cmp will invoke `source.complete`
> 2. `table.in|`
>    
>    * nvim-cmp does not invoke `source.complete` because `keyword_pattern` is `\w` typically.

Thank you, not sure how I missed that in the spec but the info was really helpful.

PR #1094, opened 2022-07-26
Fix completion menu appearance when directly below virtual text lines

tzachar:
The current implementation of `api.get_screen_cursor` does not work when completion is started right below a line of virtual text (when the entire line is virtual text). The returned window row is off by one, resulting in the completion menu hiding the current row. See the following:

b4

The fix is to use `vim.fn.winline` and `vim.fn.wincol` directly.
This is after the fix:

after

### DIFF
--- a/lua/cmp/utils/api.lua
+++ b/lua/cmp/utils/api.lua
@@ -57,9 +57,7 @@ api.get_screen_cursor = function()
     local cursor = api.get_cursor()
     return { cursor[1], cursor[2] + 1 }
   end
-  local cursor = api.get_cursor()
-  local pos = vim.fn.screenpos(0, cursor[1], cursor[2] + 1)
-  return { pos.row, pos.col - 1 }
+  return { vim.fn.winline(), vim.fn.wincol() - 1 }
 end
 
 api.get_cursor_before_line = function()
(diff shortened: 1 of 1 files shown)


Shougo:
`screenpos()` has same problem?
And `winline()` is broken when you have split same windows.

tzachar:
@Shougo `screenpos()` was used originally. It is the source of the problem. Maybe its a bug with the internal implementation of nvim.
Anyway, I did not have any problems with splits. Can you give a specific example?

Shougo:
> Anyway, I did not have any problems with splits. Can you give a specific example?

I have used `winline()` in `pum.vim`.  But if split windows, it returns another window pos.

Please see the screenshot.

スクリーンショット_2021-12-05_22-14-01

Note:  `screenpos()` returns right pos.

tzachar:
I see what you are talking about.
However, I don't have any idea how to otherwise solve this.
Maybe we should open an issue in nvim?

tzachar:
Opened an issue in nvim upstream: neovim/neovim#19543

tzachar:
Looks like the upstream issue is getting no traction, and I have no idea how to debug the problem myself. I changed the pull request to solve @Shougo 's problem, such that I am explicitly working around the screenpos bug.

tzachar:
This has been fixed upstream

PR #1117, opened 2022-08-06
add docs view filetype

askfiy:
As you can see, I have added file types for docs-view.

This will be more friendly to some plugins, such as: [nvim-colorizer.lua](https://github.com/norcalli/nvim-colorizer.lua)

When there is no file type, no color is displayed:

With the file type:

### DIFF
--- a/lua/cmp/view/docs_view.lua
+++ b/lua/cmp/view/docs_view.lua
@@ -16,6 +16,7 @@ docs_view.new = function()
   self.window:option('linebreak', true)
   self.window:option('scrolloff', 0)
   self.window:option('wrap', true)
+  self.window:buffer_option("filetype", "cmp_documentation")
   return self
 end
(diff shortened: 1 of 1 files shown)


lua/cmp/view/docs_view.lua
```diff
@@ -16,6 +16,7 @@ docs_view.new = function()
   self.window:option('linebreak', true)
   self.window:option('scrolloff', 0)
   self.window:option('wrap', true)
+  self.window:buffer_option("filetype", "cmp-docs-view")
```
hrsh7th, line 19: Could you change it as `cmp_documentation`?

askfiy: Changes are complete

hrsh7th:
LGTM. Thank you.

PR #1130, opened 2022-08-14
WIP: Separate view implementation and configuration

hrsh7th:
- Remove wildmenu view from this repo.
- Remove ghost_menu view from this repo.
- Remove `config.window` and modify `config.view`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,10 +71,8 @@ lua <<EOF
         -- vim.fn["UltiSnips#Anon"](args.body) -- For `ultisnips` users.
       end,
     },
-    window = {
-      -- completion = cmp.config.window.bordered(),
-      -- documentation = cmp.config.window.bordered(),
-    },
+    view = require('cmp-default-view')({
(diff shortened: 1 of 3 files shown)


Shougo:
> Remove wildmenu view from this repo.
> Remove ghost_menu view from this repo.

All views split out to another plugin?

hrsh7th:
No. Perhaps just leave the default menu view.

```sh
lua/
  cmp/
  cmp-default-view/
```

tzachar:
Why remove the views? do you plan on supporting adding them back in a plugin repo?

dansalvato:
I think wildmenu is excellent in combination with ghost_text, because 90% of the time I just want to use the first completion, and I don't want the full menu popping up every time I type. I have a mapping to manually show the custom view.

If you're removing wildmenu, would you consider adding a small bit to the documentation in the Develop section about how other developers can create their own views for cmp?

Issue #972, opened 2022-05-06
Default highlight config and ColorScheme autocmd


pschyska:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/15f08a8faa22d52480cdcb9ef9ca698120f04363/doc/cmp.txt#L616) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/hrsh7th/nvim-cmp/issues) and there are no open or closed issues with the same problem.

### Neovim Version

d9ec57e16a13f66a4b17edd872786e2c67348752

### Minimal reproducible full config

Couldn't get it reproduced outside of my config.

### Description

Hey,
I'm using https://github.com/RRethy/nvim-base16, and set up the colorscheme like this:

```lua
local base16 = require("base16-colorscheme")
base16.setup("ocean")
```

This does not trigger the  ColorScheme autocommand, so that nvim-cmp doesn't set the right default colors.

I can trigger it manually, but the highlights are still wrong:
```lua
if vim.env.TERM ~= "linux" then
  local base16 = require("base16-colorscheme")
  base16.setup("ocean")
  -- use :colorscheme instead of setup() to trigger cmp's autocommand
  -- cmd("colorscheme base16-ocean")
  vim.api.nvim_set_hl(
    0, "TermCursorNC", {fg = base16.colors.base00, bg = base16.colors.base00}
  )

  vim.api.nvim_set_hl(0, "CmpItemKind", {fg = base16.colors.base09})
end
-- re-trigger ColorScheme
local autocmd = require("cmp.utils.autocmd")
autocmd.emit("ColorScheme")
```

```
:hi CmpItemMenuDefault
CmpItemMenuDefault xxx cterm= gui= guifg=#000000
```
It should be:
```
CmpItemMenuDefault xxx guifg=#c0c5ce
```

I think that happens because `highlight.inherit('CmpItemMenuDefault', 'Pmenu', { bg = 'NONE', default = true })` is used. The default=true leading to CmpItemMenuDefault never getting updated.
Shouldn't CmpItemMenuDefault be default=false, and just `vim.api.nvim_set_hl(0, 'CmpItemMenu', { link = 'CmpItemMenuDefault', default = true })` be default=true? I.e. when the user manually updates CmpItemMenu, the highlight hook shouldn't overwrite that, but when the user changes the colorscheme, CmpItemMenu**Default** should be updated.
[...]

PR #1134, opened 2022-08-17
fix(plugin): overriding colorschemes

Iron-E:
Commit 8ce0832a5c409644fa06f34337b74f11aaa871c7 via #972 broke my colorscheme and other colorschemes which provide nvim-cmp highlights (e.g. linking `CmpItemKindText` to `Text` doesn't work). Plugins in general should definitely use `default = true` to maintain compatability with colorschemes, otherwise they will clobber the groups set by the colorscheme (which is what is happening now).

I looked at #972, and it seems like that issue can be fixed by delaying the setup of highlight groups until they are first needed (e.g. when rendering the popup menu), which would give that user time to setup their highlight groups _and_ still adhere to `default = true`. It seems setting everything up in `plugin/cmp.lua` doesn't guarantee the user's colorscheme is fully configured.

### DIFF
--- a/plugin/cmp.lua
+++ b/plugin/cmp.lua
@@ -26,15 +26,15 @@ for kind in pairs(types.lsp.CompletionItemKind) do
 end
 
 autocmd.subscribe('ColorScheme', function()
-  highlight.inherit('CmpItemAbbrDefault', 'Pmenu', { bg = 'NONE', default = false })
-  highlight.inherit('CmpItemAbbrDeprecatedDefault', 'Comment', { bg = 'NONE', default = false })
-  highlight.inherit('CmpItemAbbrMatchDefault', 'Pmenu', { bg = 'NONE', default = false })
-  highlight.inherit('CmpItemAbbrMatchFuzzyDefault', 'Pmenu', { bg = 'NONE', default = false })
-  highlight.inherit('CmpItemKindDefault', 'Special', { bg = 'NONE', default = false })
-  highlight.inherit('CmpItemMenuDefault', 'Pmenu', { bg = 'NONE', default = false })
+  highlight.inherit('CmpItemAbbrDefault', 'Pmenu', { bg = 'NONE', default = true })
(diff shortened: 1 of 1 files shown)


hrsh7th:
@Iron-E I want to check your colorscheme definition. If you orverride `Cmp*Default` colors. I think it's wrong. You should use `Cmp*` colors.

Iron-E:
Thanks for following up @hrsh7th. I have been using the [`CmpDefault` groups](https://github.com/Iron-E/nvim-highlite/blob/master/colors/highlite.lua#L936-L964), for two reasons:

1. Currently each `Cmp*Default` group only has one inheritor (e.g. `CmpItemKindVariable` to `CmpItemKindVariableDefault`), however in the case that more are added in the future (e.g. `hi def link CmpItemKindIdentifier CmpItemKindVariableDefault`) I wanted the definitions to be forward-compatible 
2. Other plugins or configurations may (potentially by mistake, as you mentioned) reference the `Cmp*Default` groups when deriving their own highlight groups, rather than the equivalent `Cmp*` highlight group (e.g. `hi def link MyGroup CmpItemKindVariableDefault`) which would cause mishighlighting. Defining the default groups in my colorscheme prevents any issues like this from ever arising

Admittedly this may be an unintended use of these highlight groups, although I think these are good reasons to prefer providing definitions for those groups (i.e. forward-compatibility, preventing a certain class of bug for users of my colorscheme)

hrsh7th:
The `CmpItemKind%KIND_NAME%` highlight definition is already care about your intention.

The inheritance is the following so I don't think you don't worry about `forward-compatible`.

`CmpItemKind%KIND_NAME%` -> `CmpItemKind` -> `CmpItemKindDefault`

The colorscheme can override `CmpItemKind` directly then `CmpItemKind%KIND_NAME%` will be followed.

We shouldn't care about second point. I think should fix the wrong definition.

Iron-E:
Alright, good enough for me.

PR #1143, opened 2022-08-20
Fix the complete in insert mode

FallenAngel97:
Fixes the complete in the command line of neovim

Without it:

image

With this:

image


FallenAngel97:
Sorry, I see, that one hour ago it was fixed. Don't bother :)

PR #1140, opened 2022-08-19
fix(entry/get_replace_range): handle TextEdit.range

lvimuser:
fixes cmp.mapping.confirm when ConfirmBehavior is set to Replace

----

Some servers send a `TextEdit` in the following form
```
{
  newText = '"javascript.suggest.names": ${1:true}',
  range = {
    end = {
      character = 0,
      line = 2
    },
    start = {
      character = 0,
      line = 2
    }
  }
}
```
Before/after:

https://user-images.githubusercontent.com/109605931/185707880-4819cc3e-1408-4748-940d-845f22a27073.mp4

I've seen this behavior on cssls/html/jsonls/gopls.

See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textEdit and https://github.com/LunarVim/LunarVim/issues/2876#issuecomment-1204393574.

Repro:

<details>

vimrc.vim
```vimscript
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = cmp.mapping.preset.insert{
    ["<CR>"] = cmp.mapping.confirm({ select = true, behavior = cmp.ConfirmBehavior.Replace }),
    ["<C-space>"] = cmp.mapping(function()
      cmp.complete(cmp.ContextReason.Manual)
    end),
  },

  sources = cmp.config.sources({
    { name = "nvim_lsp" },
  }),
}
EOF

lua << EOF
[...]

PR #1148, opened 2022-08-22
Fix ConfirmBehavior.Replace

keithpl:
This reverts breaking change in 27bc575.

Before this change confirm() with ConfirmBehavior.Replace behaved as
expected:

abcd|<C-n><CR>
abcdefg|

Presently confirm() with ConfirmBehavior.Replace is inserting the entire
suggestion after the cursor:

abcd|<C-n><CR>
abcdabcdefg|

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -333,12 +333,9 @@ end
 entry.get_replace_range = function(self)
   return self.cache:ensure({ 'get_replace_range', self.resolved_completion_item and 1 or 0 }, function()
     local replace_range
-    if misc.safe(self:get_completion_item().textEdit) then
-      if misc.safe(self:get_completion_item().textEdit.replace) then
+    if misc.safe(self:get_completion_item().textEdit) and
+        misc.safe(self:get_completion_item().textEdit.replace) then
         replace_range = self:get_completion_item().textEdit.replace
-      else
-        replace_range = self:get_completion_item().textEdit.range
-      end
     else
       replace_range = {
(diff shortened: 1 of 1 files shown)


keithpl:
This is the current behavior:
https://user-images.githubusercontent.com/682595/186026832-f817c995-02a3-4735-bfc2-93f7ed9c61ef.mp4

This is after reverting:
https://user-images.githubusercontent.com/682595/186026856-fbcc15bb-4abc-41e5-8d2b-08897bcbe1bc.mp4

hrsh7th:
First, most language servers tested in VSCode. So it's importa t that the lsp related plugin works the same as the VSCode.

I think the VSCode handles the `InsertReplaceTextEdit` as the same as latest nvim-cmp.

https://github.com/microsoft/vscode/blob/main/src/vs/editor/contrib/suggest/browser/suggest.ts#L239
https://github.com/microsoft/vscode/blob/main/src/vs/editor/contrib/suggest/browser/suggest.ts#L101

keithpl:
I'm not sure what the next step is; before 27bc545 ConfirmBehavior.Replace was working as expected, post 27bc545 ConfirmBehavior.Replace started inserted the entire suggestion at the cursor position.

Are you stating the problem is with the language server?

lvimuser:
See https://github.com/hrsh7th/nvim-cmp/pull/1140 and https://github.com/hrsh7th/nvim-cmp/issues/1156
The change wasn't without reason, simply reverting it will cause a regression

PR #1155, opened 2022-08-26
Fix grammar mistake: "setup" -> "set up"

amarakon:
The use of "setup" in "Setup nvim-cmp" and "Setup lspconfig" is incorrect because "setup" is a noun while "set up" is a verb. "Set up" is correct.


hrsh7th:
Thank you~!

PR #1162, opened 2022-08-30
Add incsearch_redraw_keys settings

yuki-yano:
When using incsearch with cmp-cmdline, the completion window display may be corrupted.

By default, it uses `<Up><Down>`, but you can re-render it by using `<C-r><Esc>` or `<C-r><BS>`.
However, using `<C-r>` breaks the history going back with `<Up>` and `<Down>`.

Therefore, a configuration item has been added to allow the user to change the key to be sent.

https://user-images.githubusercontent.com/5423775/187353553-2e3ec752-3a52-43fc-9781-2663905a53aa.mp4

### DIFF
--- a/lua/cmp/utils/misc.lua
+++ b/lua/cmp/utils/misc.lua
@@ -226,12 +226,13 @@ misc.redraw = setmetatable({
   doing = false,
   force = false,
   -- We use `<Up><Down>` to redraw the screen. (Previously, We use <C-r><ESC>. it will remove the unmatches search history.)
-  termcode = vim.api.nvim_replace_termcodes('<Up><Down>', true, true, true),
+  incsearch_redraw_keys = '<Up><Down>',
 }, {
   __call = function(self, force)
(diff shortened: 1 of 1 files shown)


lua/cmp/config/default.lua
```diff
@@ -101,5 +101,7 @@ return function()
         winhighlight = 'FloatBorder:NormalFloat',
       },
     },
+
+    incseach_redraw_keys = '<Up><Down>'
```
hrsh7th, line 105: I don't think it's good to put here.

I think it should be hidden API. So `require('cmp.utils.misc').redraw.incsearch_redraw_keys = '<Up><Down>'` is better.

yuki-yano: Fixed.
I have `require('cmp.utils.misc').redraw.incsearch_redraw_keys = '<C-r><BS>'` and it works fine.

hrsh7th:
LGTM

Issue #1173, opened 2022-09-02
Is there any options can remove `~` in snippet completion window?


NEX-S:
I'm wondering how to remove `~ `in the completion window like `for ~` to `for`, I think the icon already shows us that this item is a snippet, and it's expandable, so `~` in the completion window feels a bit redundant, does that make sense?

![Screenshot from 2022-09-02 08-28-31](https://user-images.githubusercontent.com/97107165/188130322-b4ce5567-e3a2-4523-a3f7-4b5e8d6a2008.png)

I have found it in entry.lua: 238
```lua
    -- ~ indicator
    local is_snippet = false
    if #(misc.safe(completion_item.additionalTextEdits) or {}) > 0 then
      is_snippet = true
    elseif completion_item.insertTextFormat == types.lsp.InsertTextFormat.Snippet then
      is_snippet = self:get_insert_text() ~= word
    elseif completion_item.kind == types.lsp.CompletionItemKind.Snippet then
      is_snippet = true
    end
    if is_snippet then
      abbr = abbr .. "~" <------------------------ HERE
    end
```

hrsh7th:
The `~` is appended if the `item is snippet` or `item has additionalTextEdit` so it's `expandable` indicator.

I can add the `formatting.expandable_indicator = false` option. PR welcome.

PR #1175, opened 2022-09-08
Remove unnecessary tilde for snippets

amarakon:
There is no need for `~` indicator for snippets because it already says "snippet" beside the snippet. See issue #1173.

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -235,19 +235,6 @@ entry.get_vim_item = function(self, suggest_offset)
     local word = self:get_word()
     local abbr = str.oneline(completion_item.label)
 
-    -- ~ indicator
-    local is_snippet = false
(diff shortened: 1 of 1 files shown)


hrsh7th:
The `~` is shown if the completion item has additionalTextEdits.

You can remove `~` via cutsomizing the `format.formatting` function.

amarakon:
> You can remove `~` via cutsomizing the `format.formatting` function.

I could not find documentation on this function in the wiki. How can I use this function to remove the `~`?

hrsh7th:
Ah, Sorry. `formatting.format` is correct.

amarakon:
> Ah, Sorry. `formatting.format` is correct.

I read the documentation for this function in `cmp.txt` and I did not understand it. How can I use it to remove the `~`? I use LuaSnip.

PR #1177, opened 2022-09-10
fix(get_replace_range): workaround for end char matching cursor position

### LINKED ISSUE #1156: Broken replace behavior
After 27bc575d1ef83fa3260c0b56cf60f3496cfab68d commit, replace behavior no longer works with Pyright LSP server when confirming entries that have auto imports. Instead of replacing text, it behaves like insert. Without auto imports, everything works as expected.

After a little investigation, I found that "2" statement is called with auto imports and "3" without them.
```lua
---Return replace range
---@return lsp.Range|nil
entry.get_replace_range = function(self)
  return self.cache:ensure({ 'get_replace_range', self.resolved_completion_item and 1 or 0 }, function()
    local replace_range
    if misc.safe(self:get_completion_item().textEdit) then
      if misc.safe(self:get_completion_item().textEdit.replace) then
        print('1')
        replace_range = self:get_completion_item().textEdit.replace
      else
        print('2') -- Called with auto imports and behaves like insert
        replace_range = self:get_completion_item().textEdit.range
      end
    else
      print('3') -- Called without auto imports and behaves as replace
      replace_range = {
        start = {
          line = self.source_replace_range.start.line,
          character = math.min(misc.to_utfindex(self.context.cursor_line, self:get_offset()), self.source_replace_range.start.character),
        },
        ['end'] = self.source_replace_range['end'],
      }
    end
    return replace_range
  end)
end
```
After reverting this commit, the issue disappears.

lvimuser:
fixes #1156, supersedes #1148 and does not regress #1140

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -339,7 +339,9 @@ entry.get_replace_range = function(self)
       else
         replace_range = self:get_completion_item().textEdit.range
       end
-    else
+    end
+
+    if not replace_range or (self.context.cursor.character == replace_range['end'].character) then
       replace_range = {
         start = {
           line = self.source_replace_range.start.line,


lua/cmp/entry.lua
```diff
@@ -339,7 +339,9 @@ entry.get_replace_range = function(self)
       else
         replace_range = self:get_completion_item().textEdit.range
       end
-    else
+    end
+
+    if not replace_range or (self.context.cursor.character == replace_range['end'].character) then
```
hrsh7th, line 344: I understood this to mean that if the obtained replace-range looks like an insert-range, it creates a replace-range with its own logic. Is this correct?

lvimuser: > obtained replace-range looks like an insert-range

Not sure tbh, but the second part is correct.

PR #1186, opened 2022-09-24
`README.md`: use buffer source for `/` and `?`

amarakon:
Since we can use both `/` and `?` for searching in Neovim, it makes more sense to use the buffer source in both of them instead of just `/`. Since putting `{ '/', '?' }` instead of `'/'` does not work, I put it in a `for` loop.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,13 +102,15 @@ lua <<EOF
     })
   })
 
-  -- Use buffer source for `/` (if you enabled `native_menu`, this won't work anymore).
-  cmp.setup.cmdline('/', {
-    mapping = cmp.mapping.preset.cmdline(),
-    sources = {
-      { name = 'buffer' }
-    }
-  })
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM.

README.md
```diff
@@ -102,13 +102,15 @@ lua <<EOF
...
 
-  -- Use buffer source for `/` (if you enabled `native_menu`, this won't work anymore).
-  cmp.setup.cmdline('/', {
-    mapping = cmp.mapping.preset.cmdline(),
-    sources = {
-      { name = 'buffer' }
-    }
-  })
+  -- Use buffer source for `/` and `?` (if you enabled `native_menu`, this won't work anymore).
+  for _,v in pairs({ '/', '?' })
```
mmirus, line 106: @amarakon I think this needs `do` at the end: `for _,v in pairs({ '/', '?' }) do`.

amarakon: Yes, it does. I forgot to put it, that is my bad.

PR #1193, opened 2022-09-26
`setup.cmdline`: support using a table

amarakon:
I made the `cmd.setup.cmdline` function support using a string or a table as the first argument. Since it is recommended to have the same configuration for both `/` and `?`, this makes a lot of sense so we do not have to use a `for` loop. I also changed the README since we do not need the `for` loop anymore.


hrsh7th:
LGTM!

PR #1194, opened 2022-09-26
Update type annotations for setup.cmdline

yuys13:
Updated type annotations as `setup.cmdline` now accepts string[] by #1193.


hrsh7th:
LGTM.

PR #1195, opened 2022-09-26
Format by stylua

yuys13:
CI is failing on the `stylua --check`, so I formatted it with stylua.
`make integration` still fails vusted, but it didn't seem like it would be fixed right away, so I would like to PR just the format first.


hrsh7th:
LGTM. Thank you.

PR #1211, opened 2022-10-03
fix(context): `in_treesitter_capture`

f3fora:
`in_treesitter_capture` used to compare a node:type() with what should be a capture, returning always a falsy value.
The example https://github.com/hrsh7th/nvim-cmp/wiki/Advanced-techniques#disabling-completion-in-certain-contexts-such-as-comments did not work as expected, being in_treesitter_capture("comment") always not true.
Now it works as expected.

### DIFF
--- a/lua/cmp/config/context.lua
+++ b/lua/cmp/config/context.lua
@@ -21,57 +21,21 @@ end
 ---@param capture string | []string
 ---@return boolean
 context.in_treesitter_capture = function(capture)
-  local highlighter = require('vim.treesitter.highlighter')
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM

smjonas:
This is a breaking change for Neovim users that haven't upgraded to 0.8 stable yet. The reason is that `get_captures_at_position` has been renamed to `get_captures_at_pos`: https://github.com/neovim/neovim/pull/20331. Could you please add a fallback to `get_captures_at_position` in case `get_captures_at_pos` is `nil`?

PR #1222, opened 2022-10-09
Fixed conflicts with ultisnip and copilot in cmp

malisoft:
Previusly i present problems with copilot and ultisnipts in this awasome plugin cmp, the complicts appears when declare variables  with "_" so i added lines code to validate execute the callbacks to get autocompletation code suggestion

image

now

image


hrsh7th:
I think the error always should be reported. If some source raised the error, I think the fix should be applied the source side.

malisoft:
> I think the error always should be reported. If some source raised the error, I think the fix should be applied the source side.
okay, in next time

PR #1224, opened 2022-10-10
fix(context): compatibility with nvim<0.8

f3fora:
`get_captures_at_position` has been renamed to `get_captures_at_pos`: neovim/neovim#20331. Fallback to `get_captures_at_position` in case `get_captures_at_pos` is `nil`.

_See comment by @smjonas in https://github.com/hrsh7th/nvim-cmp/issues/1211#issuecomment-1273592539_

### DIFF
--- a/lua/cmp/config/context.lua
+++ b/lua/cmp/config/context.lua
@@ -33,9 +33,13 @@ context.in_treesitter_capture = function(capture)
     col = col - 1
   end
 
+  local get_captures_at_pos = -- See neovim/neovim#20331
+    require('vim.treesitter').get_captures_at_pos -- for neovim >= 0.8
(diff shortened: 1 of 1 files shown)


max397574:
you can just use vim.treesitter.* instead of require"vim.treesitter".* btw

f3fora:
I am aware. Looking at the previous code, I assumed that `require('vim.treesitter')` was a common pattern in `nvim-cmp`.

hrsh7th:
Thank you.

To be honest, I prefer `vim.treesitter` instead of `require('vim.treesitter')`.
The nvim-cmp had getting a lot of contribution. So sometimes code-style is different.

Issue #1213, opened 2022-10-03
mapping.confirm doesn't escape correctly


figsoda:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```lua
local cmp = require("cmp")

vim.keymap.set("i", "<cr>", function() return "<cr>" end, { expr = true })

cmp.setup({
  mapping = {
    ["<cr>"] = cmp.mapping.confirm(), -- works without this line
  },
})
```

### Description

`mapping.confirm` doesn't escape correctly. Pressing enter types out `<cr>` instead of a new line

### Steps to reproduce

Open a file, press enter
```
a|b
```

### Expected behavior

```
a
b
```

### Actual behavior

```
a<cr>b
```

### Additional context

probably caused by breaking changes in nvim v0.8

also see https://github.com/neovim/neovim/pull/19598

hrsh7th:
Thank you. I've fixed it.

figsoda:
Thanks for the fix. Do you mind updating it on luarocks?

hrsh7th:
I can't understand what updating luarocks. But feel free to do.

briandipalma:
This seems to break nvim 0.7.

appelgriebsch:
Yes, same here. As soon as an <CR> is pressed it gives an error trace....
```
E5108: Error executing lua: ...site/pack/packer/start/nvim-cmp/lua/cmp/utils/keymap.lua:245: invalid key: replace_keycodes
stack traceback:
        [C]: in function 'nvim_set_keymap'
        ...site/pack/packer/start/nvim-cmp/lua/cmp/utils/keymap.lua:245: in function 'set_map'
        ...site/pack/packer/start/nvim-cmp/lua/cmp/utils/keymap.lua:131: in function 'callback'
        ...m/site/pack/packer/start/nvim-cmp/lua/cmp/utils/misc.lua:13: in function 'fallback'
        ...te/pack/packer/start/nvim-cmp/lua/cmp/config/mapping.lua:181: in function 'on_keymap'
        ...re/nvim/site/pack/packer/start/nvim-cmp/lua/cmp/core.lua:145: in function 'callback'
[...]

PR #1227, opened 2022-10-11
fix compatibility with nvim < 0.8

figsoda:
see comments on #1213

cc @briandipalma @appelgriebsch


hrsh7th:
Ph. I didn't know how to nvim's version checking. Thank you!

Bohun9:
Why this bug was not fixed for nvim 0.8?

hrsh7th:
What do you mean?
nvim v0.8 has breaking change. This patch is checking nvim version.

PR #1237, opened 2022-10-15
Fix undefined new() error in example source

nat-418:
Ran into a bug in the help file showing an example source. There was a stray `source.new()` call that threw errors. This PR might save somebody in the future a bit of head-scratching.


hrsh7th:
LGTM. Thank you!

PR #1293, opened 2022-11-10
fix: "count" option inserting characters

### LINKED ISSUE #1291: selecting entries adds a 1 every time an item is selected
### FAQ

### Announcement

```vim

```

### Description

When selecting a completion item, a "1" gets added every time i select an item.

Screenshot_20221110_140458

This seems to be related to the changes in https://github.com/hrsh7th/nvim-cmp/commit/bdfadc1238ecfc3e2d99fabc119e75dc575af4a1.

select_next_item() also seems to require `{ behavior = cmp.SelectBehavior, count = 1 }` now, otherwise i get:
```
...te/pack/packer/start/nvim-cmp/lua/cmp/utils/feedkeys.lua:10: bad argument #1 to 'match' (string expected, got nil)
```

Here's my conf: https://github.com/abenz1267/nvim/blob/master/lua/editing/cmp.lua

### Steps to reproduce

Just trying to autocomplete something

### Expected behavior

not erroring or adding 1s to the buffer.

### Actual behavior

erroring/adding 1s to the buffer.

### Additional context

Reverting back to the commit before that change fixes the problem.

stnley:
Closes #1291

### DIFF
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -140,9 +140,15 @@ cmp.select_next_item = cmp.sync(function(option)
     return true
   elseif vim.fn.pumvisible() == 1 then
     if option.behavior == cmp.SelectBehavior.Insert then
-      feedkeys.call(keymap.t(string.format('%s<C-n>', option.count)), 'in')
+      while option.count > 0 do
+        feedkeys.call(keymap.t('<C-n>'), 'in')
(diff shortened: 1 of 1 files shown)


stnley:
I'm sure this is not the most elegant solution but it works :man_shrugging:

PR #1299, opened 2022-11-12
fix: type annotation for enabled field in ConfigSchema

figsoda:
I was using it with https://github.com/sumneko/lua-language-server, and it was giving me a type error with `enabled = true`, probably because it parsed the type as `fun(): (boolean|boolean)`, which is equivalent to `fun(): boolean`
changing it to `boolean | fun(): boolean` fixes it for me


figsoda:
ci timed out in the `setup luarocks` stage, not sure how it happened

Issue #1087, opened 2022-07-23
[Question]: Disable Scrollbar


techtycho:
How to hide the scrollbar, because it looks very bad in my config:

![2022-07-23_16-19](https://user-images.githubusercontent.com/79283282/180609188-26935fce-1236-4537-b2b3-141d5c52f1db.png)

Can I hide the scrollbar, I want to still be able to scroll with keybindings, but with a hidden scrollbar.

ratheesh:
It would be really nice to bring back the `thin scrollbar` that was introduced during the development of floating window based completion window but later removed. 
`thin scrollbar` looks slick!

techtycho:
Hey, can I even change the scrollbar characters?

ratheesh:
@techtycho, We had this option during the development of the the floating window based completion PR. But, it was removed before integration into the main branch.

techtycho:
Is the scrollbar related to `nvim-cmp` specifically, or just a `Neovim API` feature?

Shougo:
The scrollbar is the original feature in nvim-cmp.

techtycho:
Is there anything I can do to get it right? At least make it look good on my config instead of getting rid of it?

Alex-vZyl:
Any news on this? :smile:

DasGandlaf:
Sorry, I also need this...

barrett-ruth:
@Alex-vZyl I just deleted the if statement on my copy of nvim-cmp. All seems to work for me. I guess we just wait if this is every implemented but it should be pretty trivial if dodging the if block is the solution.

pure-bliss:
https://github.com/hrsh7th/nvim-cmp/pull/1308 should answer this:

```lua
cmp.setup({
  window = {
    completion = {
      scrollbar = false,
    },
  },
})
```

PR #1308, opened 2022-11-16
feat: added hide_scrollbar option to window.completition ( #1087)

pure-bliss:
Added configuration option to disable competition scrollbar.

Inspiration: [https://github.com/hrsh7th/nvim-cmp/issues/1087#issuecomment-1193701228](https://github.com/hrsh7th/nvim-cmp/issues/1087#issuecomment-1193701228)

Issue: [https://github.com/hrsh7th/nvim-cmp/issues/1087](https://github.com/hrsh7th/nvim-cmp/issues/1087)

Before:
image

After:
image

Please let me know if I missed anything.

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -630,6 +630,11 @@ window.completion.side_padding~
   `number`
   The ammount of padding to add on the completion window's sides
 
+                                     *cmp-config.window.completion.hide_scrollbar*
+window.completion.hide_scrollbar~
+  `boolean`
+  Wether the scrollbar should be hidden even if there are more items that fit
+
                                      *cmp-config.window.documentation.max_width*
 window.documentation.max_width~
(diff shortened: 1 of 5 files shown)


wookayin:
Looks nice. A suggestion: how about renaming to `scrollbar = true` instead of `hide_scrollbar = false` (the values are default option). This will allow a potential extension/configuration about scrollbar in the future without changing/breaking config fields.

pure-bliss:
Good idea. Updated

pure-bliss:
`scrollbar = opts.scrollbar or true,`: this feels like a bad idea. I never really used lua, but feel like `opts.scrollbar` will be always true. I couldn't find usages of `cmp.config.window`, so I'm not sure how could I test this, or what the correct solution here would be

PR #1314, opened 2022-11-19
feat: `enabled` for sources

ssiyad:
This PR adds `enabled` for individual sources.

This option can be used to toggle source based on context/conditions (https://github.com/hrsh7th/nvim-cmp/issues/806). I wanted `dictionary` to be enabled only in comments and, I was able to achieve that by

```lua
enabled = function (ctx)
    return context.in_treesitter_capture('comment')
end
```

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -540,6 +540,11 @@ sources[n].max_item_count~
   `number`
   The source-specific item count.
 
+                                          *cmp-config.sources[n].enabled*
+sources[n].enabled~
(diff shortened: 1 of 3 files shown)


Arrow-x:
this works great! why is taking so long to merge?

MariaSolOs:
Is there anything blocking this PR?

PeterCardenas:
@hrsh7th anyways we can get this PR through?

ssiyad:
I would like to. However, I don't use this plugin anymore. It'd be great if somebody else picked it up.

PR #1362, opened 2022-12-20
Hide snippets from autocomplete popup

sreea1408:
This commit allows snippets to be hidden from the autocomplete popup by setting hide_snippets to true in the configuration.

Note: The solution was inspired from the following discussion #759

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -29,6 +29,7 @@ return function()
       expand = function(_)
         error('snippet engine is not configured.')
       end,
+      hide_snippets = false,
     },
 
     completion = {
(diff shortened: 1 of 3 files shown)


hrsh7th:
Sorry. I don't want to add a lot of new options.

You can use `entry_filter` for this purpose.

```lua

local function entry_filter(entry)
  return entry:get_kind() ~= require('cmp').lsp.CompletionItemKind.Snippet
end

cmp.setup {
  ...
  sources = {
    {
      name = "...",
      entry_filter = entry_filter
    }
  },
  ...
}

```

tf42src:
Would be nice to have this. Have tried everything I found so far and could not disable snippets. Using lazyvim and all examples don't match.

PR #1383, opened 2022-12-30
Add `override` option

hrsh7th:
This PR adds the `sources[n].override` option that enables to override custom source behavior.

For example, you can change `is_available` condition like this.

```lua
cmp.setup {

  sources = {
    {
      name = 'nvim_lsp',
      override = {
        is_available = function()
          return context.in_treesitter_capture('comment')
        end
      }
    }
  }

}
```

This PR has following breaking changes.

- Remove `config.sources[n].keyword_pattern` option.
- Remove `config.sources[n].trigger_characters` option.

### DIFF
--- a/lua/cmp/init_spec.lua
+++ b/lua/cmp/init_spec.lua
@@ -10,7 +10,7 @@ local Keymap = require('cmp.kit.Vim.Keymap')
 describe('cmp', function()
   before_each(spec.before)
 
-  local function setup(override)
+  local function setup(override, option)
     local c = core.new()
     local s = source.new('spec', {
       complete = function(_, _, callback)
@@ -20,10 +20,10 @@ describe('cmp', function()
     c:register_source(s)
     config.set_buffer({
       sources = {
-        {
+        vim.tbl_deep_extend('keep', {
           name = 'spec',
           override = override,
-        },
+        }, option or {}),
       },
     }, vim.api.nvim_get_current_buf())
     c:prepare()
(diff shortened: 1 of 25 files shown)


max397574:
just wondering
what else does this pr add?
because I can't imagine that all those 5.6k loc are for the `override` option only

hrsh7th:
https://github.com/hrsh7th/nvim-kit

This PR adds the nvim-kit.

wookayin:
> because I can't imagine that all those 5.6k loc are for the override option only

+1, I think adding nvim-kit should be a separate commit/PR (actually I'm not sure this is good to be merged to nvim-cmp core). Why not as a separate library or dependencies? I see many duplicates with plenary.nvim. It also seems a few more refactoring would be needed. It'd be greatly appreciated if @hrsh7th can please explain more rationale and benefits of having that merged to core.

hrsh7th:
The reason for embedding nvim-kit in core is that avoiding breaking changes and new dependencies for users.

hrsh7th:
I already use nvim-kit in some of the my codebases.

https://github.com/hrsh7th/cmp-core-example
https://github.com/hrsh7th/nvim-linkedit
https://github.com/hrsh7th/nvim-gtd
https://github.com/hrsh7th/nvim-insx

As the neovim core provides higher level APIs, it becomes less necessary, but is still needed at this point.

hrsh7th:
@wookayin I don't think it's bad to add the nvim-kit to the nvim-cmp. Why do you think it's bad?

hrsh7th:
@wookayin ping

hrsh7th:
@wookayin Hi, Shouldn't I install nvim-kit in nvim-cmp?

f3fora:
Why not split the PR? One for `override` option and another for `nvim-kit`?

lua/cmp/init_spec.lua
```diff
@@ -0,0 +1,158 @@
+local async = require('cmp.utils.async')
```
hrsh7th, line 1: use nvim-kit in spec.

hrsh7th:
yes. I can't take time at the moment, and my motivation has decreased, so I closed it.

PR #1389, opened 2023-01-02
Feat: scrollbar option

futsuuu:
Added options to customize the appearance and position of the scrollbars.

For example:

```lua
scrollbar = true,
-- or
scrollbar = {
  thumb_char = " ",
  position = "edge",
},
```

image

---

```lua
scrollbar = {
  thumb_char = "▐",
  position = "inside",
},

-- change hl
```

image

---

```lua
scrollbar = {
  thumb_char = "│",
  position = "edge",
},

-- change hl
```

image

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -632,8 +632,9 @@ window.completion.side_padding~
 
                                      *cmp-config.window.completion.scrollbar*
 window.completion.scrollbar~
-  `boolean`
-  Wether the scrollbar should be enabled if there are more items that fit
+  `boolean | cmp.ScrollbarOption`
+  Wether the scrollbar should be enabled if there are more items that fit.
+  If set to `true`, is is the same as `{ position = 'edge', thumb_char = ' ' }`.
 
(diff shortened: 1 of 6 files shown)


hrsh7th:
I'll test this PR.

futsuuu:
Thanks.

However, there is a problem at the moment, the scrollbar and the text in the floatwin are covered. So I'm trying to change it so that it changes the width of the floatwin without moving the scrollbars now.

ratheesh:
@futsuuu, Thanks for this PR.
During my usage, I sometimes found discontinuation of scrollbar char like in the image below

image

This does not occur always. Kindly check this

ratheesh:
@futsuuu, May I know why was this PR closed?

futsuuu:
Sorry, but this implementation has several bugs, some of which I can't figure out the cause.
In particular, I think the text wrapping when `position = "inner"` can only be solved with a major internal change.
So, although the functionality to be added will be smaller than this, I will create another PR when I can make a stable one.

Issue #1413, opened 2023-01-24
nvim-cmp triggers 'OptionSet' event


tummetott:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/vim-vsnip'
Plug 'hrsh7th/cmp-cmdline',
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = {
    ['<CR>'] = cmp.mapping.confirm({ select = true })
  },

  sources = cmp.config.sources({
    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}
cmp.setup.cmdline(':', {
    sources = cmp.config.sources({
        { name = "cmdline" }
    })
})

local autocmd = vim.api.nvim_create_autocmd
local augroup = vim.api.nvim_create_augroup

local group = augroup('Test', { clear = true })
autocmd('OptionSet', {
    pattern = 'cursorline',
    callback = function()
        print('set cursorline option triggered')
    end,
    group = group,
})

EOF
```

### Description

Hello, 

I registered an autocmd on the `OptionSet` event with `pattern = "cursorline"` and noticed that `nvim-cmp` triggers that event quite often when using the `cmdline` module. The expected behaviour is that no events are triggered when setting an option.

[...]

PR #1415, opened 2023-01-26
Setting local window options does not trigger the "OptionSet" event

tummetott:
This PR solves #1413

I found a better way then using the `:noautocmd` ex-command. Temporarily setting `eventignore` solves the issue


hrsh7th:
LGTM. Thank you!

PR #1417, opened 2023-01-26
Improved: ALL option settings do not fire the `OptionSet` event

tummetott:
Sorry for the git crazyness, im still learning and not very comfortable with rebasing etc.

Anyway: I reverted the last commit because I missed some nvim_{win,buf}_set_option commands. Since there are many of them, I created a wrapper function to avoid duplicate code. As of now, the OptionSet event is never triggered.

Closes https://github.com/hrsh7th/nvim-cmp/issues/1413

### DIFF
--- a/lua/cmp/utils/options.lua
+++ b/lua/cmp/utils/options.lua
@@ -1,0 +1,25 @@
+local M = {}
+
+-- Set window option without triggering the OptionSet event
+---@param window number
+---@param name string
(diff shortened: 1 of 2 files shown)


tummetott:
Somethnig wrong with this PR?

hrsh7th:
No. Sorry. I missed,this PR. LGTM.

PR #1420, opened 2023-02-01
Fix fuzzy matcher adding invalid matches (#909)

### LINKED ISSUE #909: `col value outside range` with wildmenu
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/15f08a8faa22d52480cdcb9ef9ca698120f04363/doc/cmp.txt#L616) and it didn't resolve my problem.

### Issues

### Neovim Version

NVIM v0.8.0-dev+nightly-1651-ga8d5a9e81

not possible ??????

### Description

When I enter `lua print(vim.fn.get_c` I get this:

### Steps to reproduce

-

### Expected behavior

no errors

### Actual behavior

errors

### Additional context

_No response_

dansalvato:
Fixes #909 and possibly other occasional issues.

The logic made it possible for non-matches to be inserted to the matches table as a supposed match, which could cause unexpected behavior when views try to render the invalid match. Adding an additional check for `matched` is enough to ensure that only valid matches are added.


hrsh7th:
@dansalvato Thanks!
Could you provide the reproduction test-case?

I want to add test-case for [this](https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/matcher_spec.lua)

dansalvato:
@hrsh7th I came up with a good, succinct way of reproducing this.

Here is my `cmp-repro.vim`:

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require'cmp'

cmp.setup({
  snippet = {
    expand = function(args)
    end,
  },
  view = {
    entries = {name = 'wildmenu'}
  },
    { name = 'nvim_lsp' },
    { name = 'vsnip' },
  }, {
    { name = 'buffer' },
  }),
})

cmp.setup({
  experimental = {
  ghost_text = true,
  },
})
EOF

lua << EOF

require'lspconfig'.pyright.setup {
  capabilities = capabilities,
}
EOF
```

Just `nvim -u cmp-repro.vim test.py` and type this:
```py
class TestClass:
    asset_
```

Once you add the underscore to `asset_` while typing, you get the exception.

dansalvato:
Any chance we can move the test-case to an issue and get this merged in the meantime? It's an extremely small change that fixes a longstanding bug but shouldn't have any side effects.

hrsh7th:
I' merge this, but it would be helpful if you could provide test cases.

PR #1442, opened 2023-02-15
README.md: don't set 'completeopt'

Prince213:
See https://github.com/hrsh7th/nvim-cmp/discussions/941.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,8 +56,6 @@ Plug 'hrsh7th/vim-vsnip'
 
 call plug#end()
 
-set completeopt=menu,menuone,noselect
-
 lua <<EOF
   -- Set up nvim-cmp.
   local cmp = require'cmp'


hrsh7th:
Hmm... nvim-cmp looks at `completeopt` in its implementation.
But I think it can be removed.

CalebDW:
I'm confused...does it need to be set or not? I have it set to:
```
vim.opt.completeopt = { 'menuone', 'longest', 'preview' }
```

hrsh7th:
I recommend to set `set completeopt=menu,menuone,noselect`.
because it makes general completion behavior.
if you hope to use native completion menu with omnifunc via `<C-x><C-o>` etc, it might be help.

Prince213:
So nvim-cmp is not related with the builtin [ins-completion](https://neovim.io/doc/user/insert.html#ins-completion), but can be used from it, and ['completeopt'](https://neovim.io/doc/user/options.html#'completeopt') is only for that specific use case, which means it's not needed otherwise.

If that's the case, maybe this could be mentioned in README.md.

PR #1445, opened 2023-02-16
Fix: commandline ! doesn't show with <Tab> or <S-Tab>

### LINKED ISSUE #1441: ! cmdline completion is broken (regression)
### FAQ

### Announcement

```lua
-- Use cmdline & path source for ':' (if you enabled `native_menu`, this won't work anymore).
cmp.setup.cmdline(':', {
    { name = 'path' }
  }, {
    { name = 'cmdline' }
  })
})
```

### Description

Since 53f49c5145d05a53b997d3f647f97e5ac8e9bd5c was merged, shell command line completion is broken `:!` and similar (e.g. `%!` doesn't complete anymore with the available shell commands. 

### Steps to reproduce

- type `:!`
- press tab

### Expected behavior

command line completion is shown

### Actual behavior

nothing happens

### Additional context

Now that I think about it, the completion menu for commands is probably from neovim rather than the plugin, however it was very nice to have and after this regression it won't show anymore.

michelesr:
This partially reverts 53f49c5145d05a53b997d3f647f97e5ac8e9bd5c so that if you type `:!` and press `<Tab>` or `<S-Tab>` the command line completions shows up again.

Fix #1441

### DIFF
--- a/lua/cmp/config/mapping.lua
+++ b/lua/cmp/config/mapping.lua
@@ -1,5 +1,6 @@
 local types = require('cmp.types')
 local misc = require('cmp.utils.misc')
+local feedkeys = require('cmp.utils.feedkeys')
 local keymap = require('cmp.utils.keymap')
 
 local function merge_keymaps(base, override)
@@ -89,7 +90,7 @@ mapping.preset.cmdline = function(override)
         if cmp.visible() then
           cmp.select_next_item()
         else
-          cmp.complete()
+          feedkeys.call(keymap.t('<C-z>'), 'n')
         end
       end,
     },
@@ -99,7 +100,7 @@ mapping.preset.cmdline = function(override)
         if cmp.visible() then
           cmp.select_prev_item()
         else
-          cmp.complete()
+          feedkeys.call(keymap.t('<C-z>'), 'n')
         end
       end,
     },


hrsh7th:
@wuelnerdotexe  What do you think?

wuelnerdotexe:
Hi! @hrsh7th - I think the `mappings` are working as expected without any problem so far.  I think users is not getting autocompletion on `:!` because by default [cmp-cmdline](https://github.com/hrsh7th/cmp-cmdline) ignores completion for that case by default ([see here](https://github.com/hrsh7th/cmp-cmdline#ignore_cmds-string)).  I think once again that the current behavior is the most correct, because it now also respects that `ignore_cmds-string` option, which avoids user confusion as to whether the option is being taken or not.

The ideal solution would be to enable autocompletion for `:!` from [cmp-cmdline](https://github.com/hrsh7th/cmp-cmdline).

michelesr:
Any particular reason `!` and `Man` are ignored by default? It's kind of confusing, if you follow the readme from this repo, and you expect stuff to just works.

Shougo:
I think it is performance reason.
I have read the issues for that.

hrsh7th:
Yes. It's for performance.

https://github.com/hrsh7th/cmp-cmdline/issues/5
https://github.com/hrsh7th/cmp-cmdline/issues/23
https://github.com/hrsh7th/cmp-cmdline/issues/24

michelesr:
Could it be possible to add an option somewhere so that if a specific feature like `!` or `Man` is disabled it will popup the original neovim completion menu instead, rather than just doing nothing, that is also compatible with latest @wuelnerdotexe changes?

hrsh7th:
You can specify mapping by yourself.

```lua
local cmp = require('cmp')
cmp.setup.cmdline(':', {
  ...
  mapping = cmp.mapping.preset.cmdline({
    ['<Tab>'] = {
      c = function(fallback)
        if cmp.visible() then
          cmp.select_next_item()
        else
          vim.api.nvim_feedkeys(vim.api.nvim_replace_termcodes('<C-z>', true, true, true), 'ni', true)
        end
      end
    }
  })
  ...
})
```

Issue #1455, opened 2023-02-22
Checking for if vim.fn.setcmdline exists before executing it doesn't work


joebb97:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = {
    ['<CR>'] = cmp.mapping.confirm({ select = true })
  },

  sources = cmp.config.sources({
    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}


-- Use buffer source for `/` and `?` (if you enabled `native_menu`, this won't work anymore).
cmp.setup.cmdline({ '/', '?' }, {
    mapping = cmp.mapping.preset.cmdline(),
    sources = {
        { name = 'buffer' }
    }
})
EOF

lua << EOF
local capabilities = require('cmp_nvim_lsp').default_capabilities()

require'lspconfig'.cssls.setup {
  capabilities = capabilities,
}

EOF

```

### Description

https://github.com/hrsh7th/nvim-cmp/blob/d3a2833c118843859c667f64b6169e22bffe13d0/lua/cmp/view/custom_entries_view.lua#L415

[...]

PR #1456, opened 2023-02-22
fix: better check for if setcmdline() is available

joebb97:
Fix for https://github.com/hrsh7th/nvim-cmp/issues/1455


hrsh7th:
Thank you.

mburz:
@joebb97 `if(0)` is `true` in Lua. Can you please fix your check?
`if vim.fn.has('nvim-0.8')` evaluates to `true` even on versions < 0.8

Thanks!

PR #1465, opened 2023-02-26
fix: in lua if(0) evaluates to true, so check vim.fn.has == 1

joebb97:
Addresses the comment left in https://github.com/hrsh7th/nvim-cmp/issues/1455


hrsh7th:
Thank you.

Issue #1463, opened 2023-02-25
Deal with quoted candidates properly


musjj:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

<details>
<summary>init.lua</summary>

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = {
    ['<CR>'] = cmp.mapping.confirm({ select = true })
  },

  sources = cmp.config.sources({
    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}
EOF

lua << EOF
local capabilities = require('cmp_nvim_lsp').default_capabilities()

require'lspconfig'.cssls.setup {
  capabilities = capabilities,
}
EOF
```
</details>

### Description

Candidates with quotes isn't completed correctly when using autopair plugins.

### Steps to reproduce

A sample source to test with:

```lua
local source = {}

function source:complete(params, callback)
  callback({
    { label = '"foobar"' },
    { label = '"some_string"' },
  })
end
  
return source
```

### Expected behavior

`call_fn("some_str|")` -> `call_fn("some_string"|)`

### Actual behavior

`call_fn("some_str|")` -> `call_fn("some_string"|")`

[...]

PR #1466, opened 2023-02-26
fix: de-duplicate completion correctly

musjj:
When the common overlap between the candidate we're trying to complete and the characters after the cursor is less than 2, nvim-cmp will fail to de-duplicate that redundant character. For example, trying to complete `foobar` in `foob|r` will get you `foobarr`.

This fix ensures that last character of the candidate also gets checked.

Ref: https://github.com/hrsh7th/nvim-cmp/issues/1463


hrsh7th:
I'd checked the test was passed.

PR #1470, opened 2023-03-01
Use label if sortText is not set in cmp.config.compare.sort_text

davidsierradz:
According to the spec:

```
/**
 * A string that should be used when comparing this item
 * with other items. When omitted the label is used
 * as the sort text for this item.
 */
sortText?: string;
```

See https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/

### DIFF
--- a/lua/cmp/config/compare.lua
+++ b/lua/cmp/config/compare.lua
@@ -77,6 +77,13 @@ compare.sort_text = function(entry1, entry2)
     elseif diff > 0 then
       return false
     end
+  elseif entry1.completion_item.label and entry2.completion_item.label then
+    local diff = vim.stricmp(entry1.completion_item.label, entry2.completion_item.label)
+    if diff < 0 then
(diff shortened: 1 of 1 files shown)


hrsh7th:
I don't think it's correct solution.

Of these, I think that replacing `compare.length` with `compare.label` will comply with the specification. (Currently, we don't provide an implementation of compare.label, it must be defined by the user.)

nvim-cmp sorts like this:

```lua
local cmp = require('cmp')
cmp.setup {
  ...
  sorting = {
    comparators = {
      compare.offset,
      compare.exact,
      compare.score,
      compare.recently_used,
      compare.locality,
      compare.kind,
      compare.sort_text,
      compare.length,
      compare.order,
    }
  }
  ...
}
```

_google translate_

davidsierradz:
Thanks, makes sense!

PR #1475, opened 2023-03-05
fix: annotate complete option as optional

tobealive:
Annotates the complete option as optional. The current suggest default complete mapping without parameters `cmp.mapping.complete()` throws a `missing-parameter` warning if having added the cmp dir to runtime for type assistance.


hrsh7th:
Thank you~

PR #1481, opened 2023-03-09
fix(rockspec): add doc folder

Leixb:
Fixes the missing documentation was missing when installing through luarocks.


hrsh7th:
thank you!

PR #1490, opened 2023-03-18
Adds option for setting the filetype of documentation window

taDachs:
This adds the option to add a custom filetype for the documentation window. An entry with set filetype looks tile this:
```lua
    completion_item.documentation = {
      kind = require("cmp").lsp.MarkupKind.PlainText,
      ft = "ros",
      value = table.concat(completion_item.data.msg.definition, "\n"),
    }
```

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -420,7 +420,7 @@ entry.get_completion_item = function(self)
 end
 
 ---Create documentation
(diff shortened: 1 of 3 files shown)


hrsh7th:
It isn't defined in LSP spec. So I don't want to include to nvim-cmp. Sorry.

PR #1491, opened 2023-03-18
Select nth item in the completion list, and number first ten items

benlubas:
As discussed in #1482 being able to press alt + n to select the nth item in the completion list would be pretty cool. Unfortunately there is not a great way to implement that right now, and there's (as far as I could tell) no way to number the items in the completion menu. This PR is the result of that.

- Adds a configuration option to number the first 10 items in the completion list 0 through 9
- Adds `cmp.select_nth(n)` which selects the nth item in the completion list. 

I've accounted for scrolling, and the completion list being flipped.

image

The new function allows for alt+n bindings to select items like the following:
```lua
for i = 0, 9, 1 do
  local key = table.concat({ "<M-", i, ">" })
  keys[key] = function(fallback)
    if cmp.visible() and #cmp.get_entries() > i then
      return cmp.select_nth(i + 1)
    end

    return fallback()
  end
end
``` 

`select_nth` doesn't appear to work in command line completion. The numbering isn't very configurable, but I'd be happy to add some options to configure that (ie. show the numbers on the right, start at 1, number the first x items, and stuff like that)

If this gets accepted, I plan to add a wiki page detailing how to setup binds like this, with a section for macos users, b/c alt + 1 on macos by default types some symbol or another.

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -43,6 +43,7 @@ return function()
     formatting = {
       expandable_indicator = true,
       fields = { 'abbr', 'kind', 'menu' },
+      number_options = false,
       format = function(_, vim_item)
         return vim_item
       end,
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -149,6 +149,19 @@ cmp.select_next_item = cmp.sync(function(option)
   return false
 end)
 
+---Select the nth item in the completion list
+---@param n integer
+cmp.select_nth = cmp.sync(function(n, option)
+  option = option or {}
+  if cmp.core.view:visible() then
+    local release = cmp.core:suspend()
+    cmp.core.view:select_nth(n, option)
+    vim.schedule(release)
+    return true
+  end
+  return false
+end)
+
 ---Select prev item if possible
 cmp.select_prev_item = cmp.sync(function(option)
   option = option or {}
--- a/lua/cmp/types/cmp.lua
+++ b/lua/cmp/types/cmp.lua
@@ -137,6 +137,7 @@ cmp.ItemField = {
 ---@class cmp.FormattingConfig
 ---@field public fields cmp.ItemField[]
 ---@field public expandable_indicator boolean
+---@field public number_options boolean
 ---@field public format fun(entry: cmp.Entry, vim_item: vim.CompletedItem): vim.CompletedItem
 
 ---@class cmp.SnippetConfig
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -164,12 +164,14 @@ view.select_next_item = function(self, option)
   self:_get_entries_view():select_next_item(option)
 end
 
+---Select the nth menu item showing on screen
 view.select_nth = function (self, n, options)
   local entries_view = self:_get_entries_view()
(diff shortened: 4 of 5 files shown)


hrsh7th:
I think it can be supported via `formatting.format` function.
Why do you add this to the core?

hrsh7th:
I really appreciate the contribution, but I'm not really a fan of this feature. . .

Can you tell us about the superiority of this function?

benlubas:
> I think it can be supported via formatting.format function.

If you could give me an example of that, I'd love to see it. I did try that first, and couldn't come up with one that meets the following requirements: 

- Correctly labels the first 10 items 0 to 9 and nothing else
- Shows the number on the far left or far right of the completion menu
- Numbers stick to the top when the menu is scrolled
- Works when inverted

> Can you tell us about the superiority of this function?

The use case here is binding say "alt + 1/2/3/..." to select the 1st/2nd/3rd/... item in the completion menu so if you start typing, and see the item you want in the 5th slot just hitting "alt + 5" lets you immediately select it. Additionally, if you're on the 5th item, and want to jump to the 7th item, hitting "alt + 7" still works, this is in contrast to a solution which uses `cmp.select_next_item({ count = n })`.  I've also had various difficulties/bugs with the count option on `select_next_item`. This is by far the cleaner solution to the problem.

I've found bindings like this (in conjunction with the numbers) especially useful in completing text that's already in the buffer. When typing a css class like "selectable-item", there are often other words that get suggested first, but it's very easy to pick that out from a list, identify the number, and immediately select it.

benlubas:
@hrsh7th any comment here?

tmillr:
This is an interesting idea. I see the benefit, although not sure if it's a large/necessary one. I do feel like it's possible to pull this off currently (probably via a manual "hacky" solution), but not easily.

benlubas:
The non-hacky solution for me is to just a branch of my fork onto main whenever I remember. Beauty of open source :P 

Seriously though if you think there's a way to get this to work with an external plugin I'd be happy to hear it. The only way I could think to do it would be adding a floating window that shows up next to the completion menu to display the numbers and then you could setup some bindings that press `<c-n>` or `<c-p>` based on where you are in the completion menu (assuming that you can pull that information from somewhere)

but I'm not willing to implement that honestly

howarddo2208:
I would like to have this in the core, or through another plugin is also fine.

rwblokzijl:
Personally like this as it would support my usecase of `select_first_item` and `select_last_item`

kdog3682:
I would just like to add a voice that I think this is a useful feature for beginners (like myself).

The ability to press a number, and be able to immediately get a result, is really nice.

Perhaps there would be a way of integrating the feature, without disrupting what regular users are used to?

CantGetRight82:
Also very much a fan of this feature! 🔥

It is really quite good when you see the match you want, to be able to just pick it instead of figuring out which characters to type more to make it at the top.

@benlubas thank you very much for your work on this!

@hrsh7th hopefully you can reconsider adding it to the core or help on how to make a plugin for it.

benlubas:
Instead of leaving more comments, if you like this PR just give it a thumbs up please.

schester44:
Using select_next_item and `count` makes it trivial to select by index but i'm not seeing an easy way of rendering the index of each entry within the formatting function.

Does anyone know how to get the index of the entry?

PR #1499, opened 2023-03-27
fix: Use vim.treesitter.get_node_text() to avoid deprecation warning

yutkat:
fix: https://github.com/neovim/neovim/pull/22761

PR #1500, opened 2023-03-27
fix(compare): update to `vim.treesitter.get_node_text()`

wuelnerdotexe:
Signed-off-by: Wuelner Martínez <wuelner.martinez@outlook.com>


wuelnerdotexe:
Auto closed immediately because it's a duplicate of #1499 sorry.

PR #1503, opened 2023-03-29
Document cmd.ConfirmBehavior options

mjakl:
The way nvim-cmp inserts text can be controlled via the `behavior` option (part of `ConfirmOption`). There are many tutorials mentioning those, and they are well received.

Yet, the documentation does not mention them.

This PR adds a bit of documentation what options are available, and how they differ.


hrsh7th:
Thank you!

PR #1515, opened 2023-04-11
Add option for symbol matching logic.

iteratee:
The no_symbol_match makes command line completion a lot less useful. It disables any matches for file names with symbols in them. This prevents completing things like ":b foo/bar" to ":b foo/bar.txt" or ":b foo_" to ":b baz/foo_bar.txt". Add an option `disallow_symbol_nonprefix_matching` that prevents a match if it contains a symbol and isn't a prefix match. Make that option the default. Add the option to documentation and tests. Add to the examples for command line setup disabling the option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,8 @@ lua <<EOF
       { name = 'path' }
     }, {
       { name = 'cmdline' }
-    })
+    }),
+    matching = { disallow_symbol_nonprefix_matching = false }
   })
 
   -- Set up lspconfig.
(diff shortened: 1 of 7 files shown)


joshbode:
@iteratee I was just about to make a similar PR - thank you! This has been bothering me for a while :)

My other thought was to use the `keyword_pattern` per source in place of `char.is_symbol`, which for `cmp-cmdline` would include all non-blank characters:
  https://github.com/hrsh7th/nvim-cmp/compare/main...joshbode:nvim-cmp:joshbode/keyword-symbol-check

hrsh7th:
LGTM. Thank you.
Sorry. I produced conflicts unexpectedly. Could you solve it?

iteratee:
Sorry for the delay. Rebased.

Issue #1533, opened 2023-04-24
completion on the whitescope with copliot-cmp result in a broken experience


kulabun:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'zbirenbaum/copilot.lua'
Plug 'zbirenbaum/copilot-cmp'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" vim.cmd([[set completeopt=menu,menuone,noselect]])

" Setup global configuration. More on configuration below.
lua << EOF
  local cmp = require "cmp"
  cmp.setup {
    mapping = {
      ['<CR>'] = cmp.mapping.confirm({ select = true, behavior = cmp.ConfirmBehavior.Replace })
    },

    sources = cmp.config.sources({
      { name = "copilot" },
    }),
  }
EOF

lua << EOF
  local copilot = require "copilot"
  copilot.setup({
    suggestion = { enabled = false },
    panel = { enabled = false },
  })
  local copilot_cmp = require "copilot_cmp"
  copilot_cmp.setup({})
EOF

```

### Description

When copilot is used in the beginnig of the line after indentation, resulted completion is a not what expected. It looks broken.
![](https://gcdnb.pbrd.co/images/diJjAKHCjDR4.png)
![](https://gcdnb.pbrd.co/images/Iz3TWgha3XE3.png)

It looks like this is happening because of whitespace handling in cmp and '.' keyword pattern that can trigger completion on the whitespace.

### Steps to reproduce

[...]

PR #1534, opened 2023-04-24
fix: #1533

kulabun:
1. With `string.byte(self.context.cursor_line, idx) == nil` it was evaluating to true which doesn't seem correct. But usually it happen for trailing byte, so idx returned was actually equals to `offset`.
```lua
char.is_white(string.byte(self.context.cursor_line, idx))
```
2. When all the chars are whitespaces, the actual offset returns and number of spaces, but the first element index should be +1.


hrsh7th:
Thank you very much and sorry for my late review.

LGTM!

Issue #1458, opened 2023-02-23
it un-select the first item sometimes after hitting `<Tab>`


nyngwang:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```lua
local root = vim.fn.fnamemodify("./.repro", ":p")
local G = {
  extend = function (tb1, tb2)
    return vim.tbl_deep_extend('force', {}, tb1, tb2)
  end
}

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  -- do not remove the colorscheme!
  "folke/tokyonight.nvim",
  -- add any other pugins here
  {
    'hrsh7th/nvim-cmp',
    dependencies = {
      'neovim/nvim-lspconfig',
      'hrsh7th/cmp-nvim-lsp',
      'hrsh7th/cmp-buffer',
      'hrsh7th/cmp-path',
      'hrsh7th/cmp-nvim-lua',
      'hrsh7th/cmp-vsnip',
      'hrsh7th/vim-vsnip',
      'onsails/lspkind-nvim',
    },
    config = function ()
      local cmp = require('cmp')
      cmp.setup {
        snippet = {
          expand = function (args)
            vim.fn["vsnip#anonymous"](args.body)
          end
        },
      mapping = cmp.mapping.preset.insert({
        -- <option-shift-P/N>
        ['<C-S-p>'] = cmp.mapping.scroll_docs(-4),
        ['<C-S-n>'] = cmp.mapping.scroll_docs(4),
        ['<C-o>'] = cmp.mapping.complete(), -- O for Open: This will open completion menu.
[...]

PR #1547, opened 2023-05-01
select previuos selected item on `open` (#1458)

buschco:


hrsh7th:
Fixed via #1548

Thank you~!

PR #1557, opened 2023-05-04
Added a modified=false to documentation buffer…

felsangom:
Added a modified=false to documentation buffer, after the content gets updated, so it can be removed without E89 errors on :%bd


hrsh7th:
LGTM. Thank you.

PR #1574, opened 2023-05-15
perf(source): only filter up to 200 entries and dont use the cache

folke:
This PR changes the way `get_entries` works, but semantically it's still the same as before:

- don't cache filtered entries. For Tailwindcss and others, this leads to lots of memory usage
- always calculate filtered items on the fly, but stop once desired max_entries is reached

### DIFF
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -88,24 +88,14 @@ source.get_entries = function(self, ctx)
     return {}
   end
 
-  local target_entries = (function()
-    local key = { 'get_entries', self.revision }
-    for i = ctx.cursor.col, self.offset, -1 do
-      key[3] = string.sub(ctx.cursor_before_line, 1, i)
-      local prev_entries = self.cache:get(key)
-      if prev_entries then
-        return prev_entries
-      end
-    end
-    return self.entries
-  end)()
-
   local entry_filter = self:get_entry_filter()
 
   local inputs = {}
+  ---@type cmp.Entry[]
   local entries = {}
+  local max_item_count = self:get_source_config().max_item_count or 200
   local matching_config = self:get_matching_config()
-  for _, e in ipairs(target_entries) do
+  for _, e in ipairs(self.entries) do
     local o = e:get_offset()
     if not inputs[o] then
       inputs[o] = string.sub(ctx.cursor_before_line, o)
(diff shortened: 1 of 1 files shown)


lua/cmp/source.lua
```diff
@@ -119,21 +115,15 @@ source.get_entries = function(self, ctx)
...
       if entry_filter(e, ctx) then
-        table.insert(entries, e)
+        entries[#entries + 1] = e
+        if max_item_count and #entries >= max_item_count then
+          break
+        end
       end
     end
   end
-  self.cache:set({ 'get_entries', tostring(self.revision), ctx.cursor_before_line }, entries)
```
hrsh7th, line 126: I think `get_entries` cache will reduce the CPU cost. Why do you remove this?

lua/cmp/source.lua
```diff
@@ -88,24 +90,18 @@ source.get_entries = function(self, ctx)
...
-      local prev_entries = self.cache:get(key)
-      if prev_entries then
-        return prev_entries
-      end
-    end
-    return self.entries
-  end)()
+  if self.filtered.ctx and self.filtered.ctx.id == ctx.id then
+    return self.filtered.entries
+  end
```
hrsh7th, line 95: Very good. Thank you!

lua/cmp/source.lua
folke: It leads to big memory issues with tailwind and probably others.

folke: I'll see if I can incorporate the cache again in some other way that doesn't include all the memory overhead

yioneko:
Just my little thought: current implantation of limiting entries without pre-sorting is wrong and leads to suboptimal result candidates and early break of filtering seems to conflict with this. In the theory, all the entires should be done with filtering to yield the best candidates. However, by early break we'd gain huge performance improvement because the current fuzzy matching implementation is not efficient enough for some extreme cases. This might be a choice between correctness and performance.

folke:
@yioneko the current cmp implementation already does pre-filtering and only sorts the resulting entries.

But I do agree, that it would be better to include your changes as well, so that we can remove that 200 limit.

Maybe you should also open a PR with your async changes?

lua/cmp/source.lua
folke: I believe this solution is actually optimal. When the entries are huge, we filter and break as soon as we have 200 matches.

As soon as we filter down to less than 200 matches, the cache is used and the big entries list no longer needs to be fully processed.

yioneko:
@folke What I meant is that sorting should be done before limiting to preserve entries with higher expectation, and this actually conflict with early break of breaking by `max_items_count`. The current implementaion (not including this pr) is actually wrong.

> Maybe you should also open a PR with your async changes?

My changes are pretty hacky and should not be considered for merging. Also I don't have much time to work on formalizing it.

lua/cmp/source.lua
hrsh7th: I agree that this fix improves the performance.
It also makes sense that the current nvim-cmp is already an incorrect implementation, so there is no negative behavior.

However, I doubt that this approach is the right way.

That being said, I'm not planning any drastic improvements to nvim-cmp, so I agree with efficiency first.

hrsh7th:
I can merge this for now. Do you have any work left to do?

folke:
Good to go!

I'll work on an async PR after this to further optimize the code.

hrsh7th:
Thank you~!

PR #1583, opened 2023-05-21
feat: cmp async

folke:
Made part of cmp async.

* `async.new`: starts a new coroutine that will resume at most 1ms per vim.loop tick. Returns an `Async`
* `async.yield`: when called from inside a coroutine, this will yield. Does nothing when not in an async context
* `async.wrap`: wraps a function that `async.new` for every invocation
* `Async:await`: you can either pass a callback that gets executed when the async finishes, or `nil`, to synchonously wait and get the result
* `Async:cancel`: cancels the running function on next yield
* `source.get_entries`: added `async.yield` in the main loop. This method can still be called from outside an async function, but then it will of course not automatically re-schedule on the loop when taking too long
* Added some additional checks in `source.complete` to make sure the context is still the same as the one we're completing
* updated `async.throttle` to properly deal with async functions in terms of `running` and `cancelling`

This PR removes all stuttering from taildwind (biggest offender).

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -88,7 +88,11 @@ return function()
     },
 
     view = {
-      entries = { name = 'custom', selection_order = 'top_down' },
+      entries = {
+        name = 'custom',
+        selection_order = 'top_down',
+      },
+      max_items = 100,
     },
 
     window = {
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -19,6 +19,7 @@ return function()
       debounce = 60,
       throttle = 30,
       fetching_timeout = 500,
+      async_budget = 1,
     },
 
     preselect = types.cmp.PreselectMode.Item,
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -20,6 +20,7 @@ return function()
       throttle = 30,
       fetching_timeout = 500,
       async_budget = 1,
+      max_view_entries = 200,
     },
 
     preselect = types.cmp.PreselectMode.Item,
@@ -93,7 +94,6 @@ return function()
         name = 'custom',
         selection_order = 'top_down',
       },
-      max_items = 200,
     },
 
     window = {
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -93,7 +93,7 @@ return function()
         name = 'custom',
         selection_order = 'top_down',
       },
-      max_items = 100,
+      max_items = 200,
     },
 
     window = {
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -376,7 +376,10 @@ entry.match = function(self, input, matching_config)
     }
 
     local score, matches, filter_text, _
+    local checked = {} ---@type string[]
+
     filter_text = self:get_filter_text()
+    checked[filter_text] = true
     score, matches = matcher.match(input, filter_text, option)
 
     -- Support the language server that doesn't respect VSCode's behaviors.
@@ -390,16 +393,21 @@ entry.match = function(self, input, matching_config)
           accept = accept or string.find(self:get_completion_item().textEdit.newText, prefix, 1, true)
           if accept then
             filter_text = prefix .. self:get_filter_text()
-            score, matches = matcher.match(input, filter_text, option)
+            if not checked[filter_text] then
+              checked[filter_text] = true
+              score, matches = matcher.match(input, filter_text, option)
+            end
           end
         end
       end
     end
 
-    local vim_item = self:get_vim_item(self:get_offset())
-    if filter_text ~= vim_item.abbr then
+    if score == 0 then
+      local vim_item = self:get_vim_item(self:get_offset())
       filter_text = vim_item.abbr or vim_item.word
-      _, matches = matcher.match(input, filter_text, option)
+      if not checked[filter_text] then
+        _, matches = matcher.match(input, filter_text, option)
+      end
     end
 
     return { score = score, matches = matches }
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -111,7 +111,6 @@ source.get_entries = function(self, ctx)
   local inputs = {}
   ---@type cmp.Entry[]
   local entries = {}
-  local max_item_count = self:get_source_config().max_item_count or 200
   local matching_config = self:get_matching_config()
   for _, e in ipairs(target_entries) do
     local o = e:get_offset()
@@ -128,17 +127,14 @@ source.get_entries = function(self, ctx)
 
       if entry_filter(e, ctx) then
         entries[#entries + 1] = e
-        if max_item_count and #entries >= max_item_count then
-          break
-        end
       end
     end
     async.yield()
   end
 
-  -- only save to cache, when there are no additional entries that could match the filter
-  -- This also prevents too much memory usage
-  if #entries < max_item_count then
+  -- only save to cache, when there are not too many entries
+  -- This prevents too much memory usage
+  if #entries < 200 then
     self.cache:set({ 'get_entries', tostring(self.revision), ctx.cursor_before_line }, entries)
   end
 
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -130,6 +130,9 @@ source.get_entries = function(self, ctx)
       end
     end
     async.yield()
+    if ctx.aborted then
+      async.abort()
+    end
   end
 
   -- only save to cache, when there are not too many entries
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -133,6 +133,7 @@ source.get_entries = function(self, ctx)
         end
       end
     end
+    async.yield()
   end
 
   -- only save to cache, when there are no additional entries that could match the filter
@@ -274,7 +275,7 @@ end
 ---@param ctx cmp.Context
 ---@param callback function
 ---@return boolean? Return true if not trigger completion.
-source.complete = function(self, ctx, callback)
+source.complete = async.wrap(function(self, ctx, callback)
   local offset = ctx:get_offset(self:get_keyword_pattern())
 
   -- NOTE: This implementation is nvim-cmp specific.
@@ -337,7 +338,10 @@ source.complete = function(self, ctx, callback)
       context = ctx,
       completion_context = completion_context,
     }),
-    self.complete_dedup(vim.schedule_wrap(function(response)
+    self.complete_dedup(async.wrap(function(response)
+      if self.context ~= ctx then
+        return
+      end
       ---@type lsp.CompletionResponse
       response = response or {}
 
@@ -359,6 +363,9 @@ source.complete = function(self, ctx, callback)
         end
         self.revision = self.revision + 1
         if #self:get_entries(ctx) == 0 then
+          if self.context ~= ctx then
+            return
+          end
           self.offset = old_offset
           self.entries = old_entries
           self.revision = self.revision + 1
@@ -375,7 +382,7 @@ source.complete = function(self, ctx, callback)
     end))
   )
   return true
-end
+end)
 
 ---Resolve CompletionItem
 ---@param item lsp.CompletionItem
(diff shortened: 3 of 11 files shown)


folke:
Just fixed the tests. Ready for testing by anyone that feels up for it :)

folke:
@yioneko would be great if you could also review this

lua/cmp/source.lua
```diff
@@ -359,6 +363,9 @@ source.complete = function(self, ctx, callback)
         end
         self.revision = self.revision + 1
         if #self:get_entries(ctx) == 0 then
```
yioneko, line 365: Would be better to also cancel this `get_entries` call on context changed.

yioneko:
Nice work! I believe this will completely remove any user-blocking experience.

yioneko (commented):
I've tested locally and this works fine for me.

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,100 @@ async.debounce_next_tick_by_keymap = function(callback)
...
+  for _, callback in ipairs(self.callbacks) do
+    callback(result, error)
+  end
+end
+
+function Async:_step()
+  if not self.running then
+    return
+  end
+  local start = vim.loop.hrtime()
```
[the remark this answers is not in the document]
folke: Good point. Done!

lua/cmp/source.lua
folke: I just tried adding the cancellation the same way you did it, but then completion no longer always gives the correct results.

It's also not needed I think with my implementation, since the `async.throttle` will stop previous invocations already, so there will always be at most one async filter/complete running.

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,100 @@ async.debounce_next_tick_by_keymap = function(callback)
...
+      break
+    end
+  end
+  vim.schedule(function()
+    self:_step()
+  end)
+end
+
+function Async:cancel()
+  self.running = false
```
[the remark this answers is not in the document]
folke: Not completely sure what you mean, but this is used by `async.throttle` to cancel existing runinning async functions.

folke:
I removed pre-filtering from `source.get_entries` and added a new config option `view.max_items`.

The resulting items in the entries view are now always correct according to the scoring.

See https://github.com/hrsh7th/nvim-cmp/pull/1583/commits/d1136e569454fe928cf3a1c31a00e7a21be6a3da

lua/cmp/source.lua
folke: I agree, that code should probably be just removed. Or just be a check on `#self.entries` instead.

@hrsh7th what is the rationale behind this?

We're talking about this line: https://github.com/hrsh7th/nvim-cmp/blob/3ac8d6cd29c74ff482d8ea47d45e5081bfc3f5ad/lua/cmp/source.lua#L361

folke: Same remark with the code here https://github.com/hrsh7th/nvim-cmp/blob/3ac8d6cd29c74ff482d8ea47d45e5081bfc3f5ad/lua/cmp/core.lua#L329

folke: @yioneko I added abort again, but using `async.abort()`, which yields `abort` and the scheduler simply stops the whole async function. Everything works as expected now. So when the source complete context is aborted, it will never finish. The issue I was having before is that check on `0` that restored the previous entries.

See changes here https://github.com/hrsh7th/nvim-cmp/pull/1583/commits/e7872be8a03146012f4d7d0b535beefe3f9f53c5

folke:
Contexts can now be aborted. Any running `source.get_entries` will also abort their async coroutine when their context is aborted. https://github.com/hrsh7th/nvim-cmp/pull/1583/commits/e7872be8a03146012f4d7d0b535beefe3f9f53c5

bluz71:
Just tested out your PR @folke.

Note, I am not authoritative since I only dabble in Tailwind occasionally, but this PR makes nvim-cmp absolutely **buttery smooth** with the Tailwind Language Server. A beautiful developer experience to code and not stall.

This PR definitely works very well alongside the previous PR; actually in my case this PR has **more** of an effect than the previous PR; though both compliment each other very well.

Two big thumbs up from me.

Great work @folke and @yioneko.

folke:
Thanks!
@bluz71 What other PR are you referring to?

bluz71:
> Thanks! @bluz71 What other PR are you referring to?

#1574, "perf(source): only filter up to 200 entries and dont use the cache"

folke:
I get what you mean now 🙂 Yes, this pr will have a bigger effect. The other pr mainly fixed a memory leak

jorgebef:
I just wanted to say that I do tend to use Tailwind a bit and was suffering from severe lag prior to these PRs.
This one improves the experience significantly. Thank you for the awesome work as usual, big fan!

The only thing I noticed in my very quick testing is that when entering insert mode and very quickly typing right after that, occasionally it would ignore the very first character that has been typed.

For example, lets say I have my cursor here in normal mode (noted by | ) : `<div className="h-2| w-3"`
And *very quickly* enter insert mode and type `flex`, resulting in: `<div className="h-2 flex|  w-3"`

It would occasionally provide results as if only `lex` was typed (left-px and others) and not the full `flex` which was actually typed, so basically ignoring the very first character.
I attribute this to maybe the initial milliseconds that it takes for cmp to launch and has nothing to do with the scope of the PR?

I was not able to reproduce constantly, but if I really tried hard, I was able to achieve this a few times.
Please excuse my ignorance on the subject, and again, this is a very fringe case, but I thought it was maybe worth mentioning.

folke:
@jorgebef just tried that and I can't seem to replicate this. Is this something that only happens with this PR? Or something you have without this PR as well?

folke:
@jorgebef I'm able to reproduce this as well, but it also happens without this PR, so it's a bug in cmp.

image

I'll see if I can fix this as well

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,118 @@ async.debounce_next_tick_by_keymap = function(callback)
   end
 end
 
+local Scheduler = {}
+Scheduler._queue = {}
+Scheduler._executor = assert(vim.loop.new_check())
+
+function Scheduler.step()
+  local start = vim.loop.hrtime()
+  while #Scheduler._queue > 0 and vim.loop.hrtime() - start < 1e6 do
```
hrsh7th, line 172: Could you convert it as a performance option?

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,118 @@ async.debounce_next_tick_by_keymap = function(callback)
...
+  if #Scheduler._queue == 0 then
+    return Scheduler._executor:stop()
+  end
+end
+
+---@param a Async
+function Scheduler.add(a)
+  table.insert(Scheduler._queue, a)
+  if not Scheduler._executor:is_active() then
+    Scheduler._executor:start(vim.schedule_wrap(Scheduler.step))
```
hrsh7th, line 188: I'm not familiar with uv_check_t.

However, I was worried that this would call `vim.schedule_wrap` less times than the `uv_check_t` start callbacks.

I've checked this. It's perfectly OK. Thank you~!

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,118 @@ async.debounce_next_tick_by_keymap = function(callback)
...
+  elseif res == 'abort' then
+    return self:_done(nil, 'abort')
+  elseif coroutine.status(self.thread) == 'dead' then
+    return self:_done(res)
+  end
+end
+
+function Async:cancel()
+  self.running = false
+end
```
hrsh7th, line 236: I think it (and `Async:abort`) should also call the `Scheduler.remove(this)`.

The `L173-174` will call `a:_step` without running check on next iteration.

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,118 @@ async.debounce_next_tick_by_keymap = function(callback)
...
+      table.insert(self.callbacks, cb)
+    else
+      cb(self.result, self.error)
+    end
+  else
+    while self.running do
+      vim.wait(10)
+    end
+    return self.error and error(self.error) or self.result
+  end
```
hrsh7th, line 250: I think the part of `vim.wait` should be separated method that should be `Async:sync`

lua/cmp/utils/async.lua
```diff
@@ -147,4 +163,118 @@ async.debounce_next_tick_by_keymap = function(callback)
...
+      return fn(unpack(args))
+    end)
+  end
+end
+
+-- This will yield when called from a coroutine
+function async.yield(...)
+  if not coroutine.isyieldable() then
+    return ...
+  end
```
hrsh7th, line 272: I think it's better to throw error for future development.

lua/cmp/source.lua
```diff
@@ -128,16 +127,17 @@ source.get_entries = function(self, ctx)
...
   end
 
-  -- only save to cache, when there are no additional entries that could match the filter
-  -- This also prevents too much memory usage
-  if #entries < max_item_count then
+  -- only save to cache, when there are not too many entries
+  -- This prevents too much memory usage
+  if #entries < 200 then
     self.cache:set({ 'get_entries', tostring(self.revision), ctx.cursor_before_line }, entries)
   end
```
hrsh7th, line 142: I'll change to be the latest only full cache later.

hrsh7th:
Thank you so much for your PR!

lua/cmp/config/default.lua
```diff
@@ -88,7 +88,11 @@ return function()
     },
 
     view = {
-      entries = { name = 'custom', selection_order = 'top_down' },
+      entries = {
+        name = 'custom',
+        selection_order = 'top_down',
+      },
+      max_items = 100,
```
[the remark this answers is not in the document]
folke: I took it out of entries, because `entries` can also be a `string`. See:

```lua
---@alias cmp.EntriesConfig cmp.CustomEntriesConfig|cmp.NativeEntriesConfig|cmp.WildmenuEntriesConfig|string
```

I can probably still move it inside though. Let me know :)

lua/cmp/utils/async.lua
folke: Yeah, checked that with bfredl first :)

lua/cmp/utils/async.lua
folke: The `Async:_step` function handles aborts. When a function is `aborted`, the `Async` object will correctly mark itself as done with an error state of `abort`

So this is what happens:
* `async.abort()` yields `abort`
* `Async._step` picks this up and sets `self.running = false`
* `Scheduler.step` sees it's no longer running and won't reschedule this `Async`

lua/cmp/source.lua
folke: What do you mean exactly? I don't fully understand your comment :)

lua/cmp/config/default.lua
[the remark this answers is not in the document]
hrsh7th: It was supposed to adopt the default-value if a string-value was specified.

However, on second thought, having the same setting in Custom/Native/Wild is redundant.

Semantically, I thought it should be in the performance section. What do you think?

lua/cmp/entry.lua
```diff
@@ -376,7 +376,10 @@ entry.match = function(self, input, matching_config)
     }
```
hrsh7th, line 376: todo: check

lua/cmp/utils/async.lua
```diff
@@ -9,6 +9,7 @@ local async = {}
 ---@field public stop function
```
hrsh7th, line 10: todo: check

lua/cmp/source.lua
hrsh7th: Currently, nvim-cmp will create cache for each user input.

1. user-input=`use`, cache-key: `get_entries:%REVISION%:use`
2. user-input=`useE`, cache-key: `get_entries:%REVISION%:useE`
3. ...

This has benefit for performance when user press the backspace key.

But now folke research shows that this cache is bad for memory.

It doesn't make sense to cache or not by the number of items here.

Either keep a cache only for the last user input, or remove caching entirely.

folke: I think I correctly implemented this in https://github.com/hrsh7th/nvim-cmp/pull/1583/commits/c0bf17c4b228e7a24975675425c15b72c06a876d

Would be great if you could double-check if the re-use conditions are 100% correct!

I also removed the `filtered.entries` from my previous PR, since that is now no longer needed.

hrsh7th:
It's perfect to me! Thank you again!

folke:
Awesome, thank you for merging!

PR #1585, opened 2023-05-21
Fix demo video in README.md

### LINKED ISSUE #1584: The video in the README fails to load
Here's what it looks like for me on Firefox 113.0.1, however I've also tried on Chromium 113.0.5672.126 (both on my linux desktop) and on iOS Safari (no idea of version) and encountered a similar issue.

videofailstoload

In the browser console for both Firefox and Chrome I see messages about a resource failing to load/being blocked due to CSP of the page.

If I navigate directly to the video [here](https://github-production-user-asset-6210df.s3.amazonaws.com/629908/139000570-3ac39587-a88b-43c6-b35e-207489719359.mp4), I'm able to view it.

neeerp:
This seems to fix #1584. (See the rich diff; the video actually shows up and can be played).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 A completion engine plugin for neovim written in Lua.
 Completion sources are installed from external repositories and "sourced".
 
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you!

PR #1592, opened 2023-05-26
fix: handle godot LSP better

### LINKED ISSUE #1119: Using the autocomplete menu removes the completion when using GDScript LSP Server
### FAQ

### Announcement

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
    ["<Tab>"] = cmp.mapping(function(fallback)
      if cmp.visible() then
        cmp.select_next_item()
      elseif vim.fn["vsnip#available"](1) == 1 then
        feedkey("<Plug>(vsnip-expand-or-jump)", "")
        cmp.complete()
      else
        fallback() -- The fallback function sends a already mapped key. In this case, it's probably `<Tab>`.
      end
    end, { "i", "s" }),

    ["<S-Tab>"] = cmp.mapping(function()
      if cmp.visible() then
        cmp.select_prev_item()
      elseif vim.fn["vsnip#jumpable"](-1) == 1 then
        feedkey("<Plug>(vsnip-jump-prev)", "")
      end
    end, { "i", "s" }),
  },

  sources = {
    { name = "nvim_lsp" },
    { name = "buffer" },
  },
}
EOF

lua << EOF

require'lspconfig'.gdscript.setup {
  capabilities = capabilities,
  }
EOF
```

### Description

If you press `<CR>` without navigating through the autocomplete menu with `<TAB>`, it will autocomplete the first selection without issues:
Peek 2022-08-06 23-25

The problem occurs when you navigate through the autocomplete menu and press `<CR>` when a item is selected:
Peek 2022-08-06 23-24

### Steps to reproduce

With the minimal config provided earlier:
1. Download Godot https://godotengine.org/download
2. Clone the repository with godot example projects https://github.com/godotengine/godot-demo-projects
3. Open the project **godot-demo-projects/2d/bullet_shower** in Godot (otherwise the GDScript LSP server will not work).
4. Open the file **godot-demo-projects/2d/bullet_shower/bullets.gd** in neovim with the minimal config provided.
5. Try to autocomplete some constant or variable with and without navigating through the autocomplete menu.

### Expected behavior

It should complete the selected item in the autocomplete menu without any issue.

### Actual behavior

It does not complete the selected item, only works when you press `<CR>` without navigating through the menu.

### Additional context

I have been digging into this problem before creating this issue, and what I have found is that the LSP server only returns a `label` without any `insert_text` in the LSP type `completionItem/resolve`.
Then the neovim lsp client or nvim-cmp creates a CompletionItem where the property `insert_text` is an empty string.

Then I found that the "problem" was in this line: https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/core.lua#L414

  If I replace that line with the next code, it solves the problem but I don't know if it can cause issues with others LSP servers or sources from nvim-cmp.
```lua
local insertText = misc.safe(completion_item.insertText)
if misc.empty(insertText) then
  insertText = nil
end
completion_item.textEdit.newText = insertText or completion_item.word or
    completion_item.label
```

I got through this solution with the LSP Spec: 
```
	/**
	 * The label of this completion item.
	 *
	 * The label property is also by default the text that
	 * is inserted when selecting this completion.
	 *
	 * If label details are provided the label itself should
	 * be an unqualified name of the completion item.
	 */
	label: string;
```
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem

My knowledge about the LSP Protocol and Lua is limited so maybe this is not correct.

Mawdac:
This fixes #1119 (basically using the solution prescribed there, except `misc.safe` is no longer a function). I'm not super familiar with LSP, but I use neovim and godot frequently and would like to see this working. Let me know how I can improve this. In my testing this fixed the issue and didn't cause any noticeable side-effects.

### DIFF
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -423,7 +423,11 @@ core.confirm = function(self, e, option, callback)
     local completion_item = misc.copy(e:get_completion_item())
     if not completion_item.textEdit then
       completion_item.textEdit = {}
-      completion_item.textEdit.newText = completion_item.insertText or completion_item.word or completion_item.label
+      local insertText = completion_item.insertText
+      if misc.empty(insertText) then
+        insertText = nil
+      end
+      completion_item.textEdit.newText = insertText or completion_item.word or completion_item.label
     end
     local behavior = option.behavior or config.get().confirmation.default_behavior
     if behavior == types.cmp.ConfirmBehavior.Replace then


hrsh7th (approved):
thank you.
it's actually server-side bug. but i can merge it :)

Mawdac:
In continuing to use this LSP I'm realizing there are many bugs with it lol.... Thanks for accepting this PR though @hrsh7th , at least myself and others (there are ten's of us!) can use it in the most basic way.

PR #1593, opened 2023-05-26
fix entry highlight in complete-menu

TwIStOy:
fix entry highlight after #1583

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -376,7 +376,7 @@ entry.match = function(self, input, matching_config)
     }
 
     local score, matches, filter_text, _
-    local checked = {} ---@type string[]
+    local checked = {} ---@type table<string, boolean>
 
     filter_text = self:get_filter_text()
     checked[filter_text] = true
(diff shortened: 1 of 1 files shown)


lua/cmp/entry.lua
```diff
@@ -402,12 +402,10 @@ entry.match = function(self, input, matching_config)
...
-      local vim_item = self:get_vim_item(self:get_offset())
-      filter_text = vim_item.abbr or vim_item.word
-      if not checked[filter_text] then
-        _, matches = matcher.match(input, filter_text, option)
-      end
+    local vim_item = self:get_vim_item(self:get_offset())
+    filter_text = vim_item.abbr or vim_item.word
+    if not checked[filter_text] then
+      _, matches = matcher.match(input, filter_text, option)
     end
```
hrsh7th, line 409: Could you change `if score ~=0` instead of removing `if score == 0`?

TwIStOy: Thanks for your review. Fixed it!

hrsh7th:
LGTM. Thank you.

PR #1596, opened 2023-05-27
feat(ghost_view): made inline ghost text optional. Enabled by default on 0.10.0

folke:
Added an `inline` config option for ghost text and a new hl group `CmpGhostText` that defaults to `Comment`.

Inline is still enabled by default on Neovim 0.10.0, but can now be disabled:

```lua
  {
    "hrsh7th/nvim-cmp",
    dev = true,
    opts = {
      experimental = {
        ghost_text = { inline = false },
      },
    },
  },
```

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -371,6 +371,9 @@ NOTE: `kind` is a symbol after each completion option.
 *CmpItemMenu*
   The menu field's highlight group.
 
+*CmpGhostText*
+  Highlight group for ghost text. Defaults to `Comment`
+
 ==============================================================================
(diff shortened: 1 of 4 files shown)


hrsh7th:
Sorry I don't want to merge this.
Because ghost_text should be inline whenever inline is available, and shouldn't need to be customizable.

I'll add feature detection.

folke:
ok, no problem. I probably just need to get used to it.
It's mostly in combination with copilot-cmp that it's sometimes weird to have ghost-text inside some code instead of only when the rest of the line is empty

PR #1607, opened 2023-06-02
Adding check for `isyieldable`.

Sammyalhashe:
I recently synced my plugins and it seems like #1583 did cause an issue. Specifically what was happening is that when getting sources, the `couroutine.isyieldable` method would not exist. I did some research looking through neovim's docs/help-pages and found no reference to this function both online and in the help-pages. My rudimentary conclusion is that potentially I am running an outdated version of neovim where this method does not exist on coroutines *yet* and I potentially need to upgrade.

If I am right in this I suggest either adding a note to the breaking changes page or refine/merge this PR so that other versions of neovim continue to work here.

My current running version of neovim is `v0.9.0 Release` with LuaJIT `2.0.4`

### DIFF
--- a/lua/cmp/utils/async.lua
+++ b/lua/cmp/utils/async.lua
@@ -276,7 +276,7 @@ end
 
 -- This will yield when called from a coroutine
 function async.yield(...)
-  if not coroutine.isyieldable() then
+  if coroutine.isyieldable ~= nil and not coroutine.isyieldable() then
     error('Trying to yield from a non-yieldable context')
     return ...
   end


hrsh7th:
I think LuaJIT supports `coroutine.isyieldable`.
https://github.com/LuaJIT/LuaJIT/blob/v2.1/doc/extensions.html#L367

Sammyalhashe:
So that explains the discrepancy. So would it be a good idea to potentially add a note in the breaking changes doc? I would find it surprising if I was the only person with an outdated LuaJit.

hrsh7th:
Yes. I agree your suggetion. We should add breaking change announcement.

but wait, what the version of the `coroutine.isyieldable` was added to LuaJIT?
I can't find any version related note from above page.

Sammyalhashe:
> Yes. I agree your suggetion. We should add breaking change announcement.
> 
> but wait, what the version of the `coroutine.isyieldable` was added to LuaJIT? I can't find any version related note from above page.

I've been having trouble finding reference to it too. In the file you linked to it seems to imply that version 5.3 is the version that it was added.

hrsh7th:
IMO, my link seems to say `LuaJIT has extention feature that's come from Lua5.3`.

elmotec:
I ran into that issue as well and being new to Lua, LuaJIT and neovim plugins in general, I am glad I found this issue because I don't know how to upgrade LuaJIT in neovim.  Thank you @hrsh7th for the plugin too.  

Question though: isn't neovim Lua tied to 5.1 though?  When I go to `:help lua-compat`, it sounds like the interface is anchored to 5.1.

lua/cmp/utils/async.lua
```diff
@@ -276,7 +276,7 @@ end
 
 -- This will yield when called from a coroutine
 function async.yield(...)
-  if not coroutine.isyieldable() then
+  if coroutine.isyieldable ~= nil and not coroutine.isyieldable() then
```
folke, line 279: This can be safely replaced by:

```lua
  if coroutine.running() == nil then
```

That way we don't need `coroutine.isyieldable`

folke:
@hrsh7th this can be closed, since I already removed `coroutine.isyieldable` in my PR you just merged :)

Issue #1608, opened 2023-06-02
Input lag after commit #abb5c7519d40314240effc70d781149c0f097af3


zaqxsw-dev:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```vim
vim.cmd [[autocmd BufWritePre *.js,*.ts,*.php,*.rs,*.cs,*.json,*.tsx,*.jsx Neoformat]]
vim.cmd [[set rtp+=~/.fzf]]
vim.cmd [[set signcolumn=yes]]

vim.opt.cmdheight = 1
vim.opt.updatetime = 50
vim.opt.tabstop = 2
vim.opt.shiftwidth = 2
vim.opt.expandtab = true
vim.opt.relativenumber = true
vim.opt.ignorecase = true
vim.opt.swapfile = false
vim.opt.backup = false
vim.opt.wrap = false
vim.opt.encoding = "utf8"
vim.opt.scrolloff = 8
vim.opt.termguicolors = true
vim.opt.smartindent = true

local core_conf_files = {
  "oldinit.vim",
}

-- source all the core config files
for _, name in ipairs(core_conf_files) do
  local path = string.format("%s/%s", vim.fn.stdpath("config"), name)
  local source_cmd = "source " .. path
  vim.cmd(source_cmd)
end

-- Plugins
require('packer').startup(function(use)
use { "wbthomason/packer.nvim" }
-- Theme
use {'xiyaowong/nvim-transparent'}
use {'ellisonleao/gruvbox.nvim'}
-- Find files and code
use {'nvim-lua/plenary.nvim'}
use {'nvim-telescope/telescope.nvim'}
use {'lewis6991/gitsigns.nvim'}
-- LSP
use {'williamboman/nvim-lsp-installer'}
use {'hrsh7th/nvim-cmp'}
use {'hrsh7th/cmp-nvim-lsp'}
use {'hrsh7th/cmp-buffer'}
use {'hrsh7th/cmp-path'}
use {'hrsh7th/cmp-cmdline'}
use {'ray-x/lsp_signature.nvim'}
use {'neovim/nvim-lspconfig'}
use {'glepnir/lspsaga.nvim', branch = 'main'}
[...]

Issue #1606, opened 2023-06-02
Block while typing after #1583


fecet:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```vim

```

### Description

I found completion become very slow after update, a git bisect tell me the first error happen at abb5c7519d40314240effc70d781149c0f097af3.

https://github.com/hrsh7th/nvim-cmp/assets/41792945/0758a705-9cd0-493d-8199-bb20627f0e8b

### Steps to reproduce

1. Open any python file
2. type a function name, like "print" or "isinstance"
3. type anything, it will block at first character

### Expected behavior

cmp will not block my typing

### Actual behavior

see above

folke:
What is needed to reproduce this? When I open a python file and start typing and don't experience any delays.

What other lsp completion related plugins are you using?

fecet:
Perhaps relavent to https://github.com/ray-x/lsp_signature.nvim and I'm using pylsp

folke:
Tried that, but still doesn't lag.

folke:
What is your lsp signature and pylsp config?

folke:
Can you give me a single python source file that triggers the issue with the exact steps to trigger it?

fecet:
Do you want a minimun config or just reproduce, is $NVIM_APPNAME ok?

folke:
minimum config, since I need to be able to properly reproduce this.

You can use this template https://github.com/folke/noice.nvim/wiki/Minimal-%60init.lua%60-to-Reproduce-an-Issue

folke:
Not sure it will fix anything, but if you could try https://github.com/hrsh7th/nvim-cmp/pull/1611 to see if that makes a difference.

There was an issue where a cancelled/stopped filter would not set `running = false`.

fecet:
I tried to extract all my relavent config but unfortunately the bug doesn't happen...It seems that someone have provided a config and you have found the reasons behind the bug, tell me if you still need a minimal config for python.

 I hope this can help you eliminate some problems:
```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
[...]

PR #1610, opened 2023-06-03
Try to fix input deplay

zaqxsw-dev:
**Try solve:** https://github.com/hrsh7th/nvim-cmp/issues/1606 https://github.com/hrsh7th/nvim-cmp/issues/1608

I'm not sure what it's correct solution. Because most important problem it this code

```lua
cmp.sync = function(callback)
  return function(...)
    cmp.core.filter:sync(1000)
    if callback then
      return callback(...)
    end
  end
end
``` 

https://github.com/hrsh7th/nvim-cmp/blob/fc0f694af1a742ada77e5b1c91ff405c746f4a26/lua/cmp/utils/async.lua#L38-L42

This function using in other cmp function like a cmp.visible. There are enough plugins using this functions.
For example for check what autocomplete menu is open lsp-signature call cmp.visible on every move cursor. If LSP server respond fastly we not see input lag. But when lsp server respond slowly we have input delay ~1 sec. IDK maybe input delay referenced with incorrect work vim.wait but more true way i think make async analogs for sync functions like cmp.visible. This will definitely solve the problem (i think)

### DIFF
--- a/lua/cmp/utils/async.lua
+++ b/lua/cmp/utils/async.lua
@@ -60,24 +60,23 @@ async.throttle = function(fn, timeout)
 
       self.running = true
       self.stop(false)
-      timer:start(math.max(1, self.timeout - (vim.loop.now() - time)), 0, function()
-        vim.schedule(function()
-          time = nil
-          local ret = fn(unpack(args))
-          if async.is_async(ret) then
-            ---@cast ret Async
-            _async = ret
-            _async:await(function(_, error)
-              self.running = false
-              if error and error ~= 'abort' then
-                vim.notify(error, vim.log.levels.ERROR)
-              end
-            end)
-          else
+
+      vim.defer_fn(function()
+        time = nil
+        local ret = fn(unpack(args))
+        if async.is_async(ret) then
+          ---@cast ret Async
+          _async = ret
+          _async:await(function(_, error)
             self.running = false
-          end
-        end)
-      end)
+            if error and error ~= 'abort' then
+              vim.notify(error, vim.log.levels.ERROR)
+            end
+          end)
+        else
+          self.running = false
+        end
+      end, math.max(1, self.timeout - (vim.loop.now() - time)))
     end,
   })
 end
--- a/lua/cmp/utils/async.lua
+++ b/lua/cmp/utils/async.lua
@@ -29,9 +29,7 @@ vim.api.nvim_create_autocmd('VimLeavePre', {
 ---@return cmp.AsyncThrottle
 async.throttle = function(fn, timeout)
   local time = nil
-  local timer = assert(vim.loop.new_timer())
   local _async = nil ---@type Async?
-  timers[#timers + 1] = timer
   return setmetatable({
     running = false,
     timeout = timeout,
@@ -44,7 +42,6 @@ async.throttle = function(fn, timeout)
       if reset_time ~= false then
         time = nil
       end
-      timer:stop()
       if _async then
         _async:cancel()
         _async = nil


zaqxsw-dev:
Yeah, i think, it's end solution. After this PR we may be get input delay only when we have slow LSP server. But how i can see this can be fixed throught performance.debounce option. Mostly the problem related with timer in throttle function, if we pass second argument > 0 we solve problem with input delay but get another problems. 
@hrsh7th @folke what do you think about this?

folke:
What's the easiest way to trigger this issue? I'm not sure your fix is a real fix for this.
Would like to look into it myself. I don't use cmp-lsp-signature. Can I trigger the issue with that?

zaqxsw-dev:
@folke Yes you can.

https://github.com/hrsh7th/nvim-cmp/blob/fc0f694af1a742ada77e5b1c91ff405c746f4a26/lua/cmp/utils/async.lua#L63-L80

Change timout(first arg) to 500-1000 and you get input lag for any lsp server.

hrsh7th:
Hm... I can't reproduce the input lag with the following debug source.

```lua
local source = {}

function source.complete(_, _, callback)
  vim.defer_fn(function()
    callback({{
      label = 'debug',
    }})
  end, 1000)
end

return source
```

folke:
@hrsh7th I can't reproduce this either. Tried a lot of different setups and even added ray lsp signature, pylsp etc. No delays for me.

I did find a possible place where `running` could be left at `true` while it was no longer running. See here https://github.com/hrsh7th/nvim-cmp/pull/1611

But would be good to have a minimal reproducible init.lua to really see what's causing these delays for some.

hrsh7th:
thank you.
lsp_signature.nvim seems to check the `cmp.visible()`. I'll check it.

hrsh7th:
Hm... I can't reproduce with your config.
(but my test project is quiet small...)

https://github.com/hrsh7th/nvim-cmp/assets/629908/4559830e-d5af-4256-9067-737d997e8fdd

hrsh7th:
I think folke's fix is essential.
Even if this issue is not fixed, I would still like to merge it.
(If we cancel the throttle, we should be able to handle it properly.)

zaqxsw-dev:
> @hrsh7th I can't reproduce this either. Tried a lot of different setups and even added ray lsp signature, pylsp etc. No delays for me.
> 
> I did find a possible place where `running` could be left at `true` while it was no longer running. See here #1611
> 
> But would be good to have a minimal reproducible init.lua to really see what's causing these delays for some.

i test you changes in my system and it not solve issue


https://github.com/hrsh7th/nvim-cmp/assets/73785768/673671cf-fd4a-4242-9efc-acf7be332f4d

I guess issue can be reproduce only on macos. What you have OS and nvim build?

https://github.com/hrsh7th/nvim-cmp/blob/b5a636d46c69bb371995c22d7a10ee1544004879/lua/cmp/utils/async.lua#L63-L80

Probably when repeat parameter is 0 this function block a thread?

folke:
To everyone here, I **still** can't reproduce  this. Can someone please provide a full minimal `init.lua` with the exact steps (including opening a file that is known to cause this) that triggers the issue?

As long as we don't have a way to fully reproduce this behavior, it's impossible to even start to understand what is causing this. Let alone fix it.

You can use the template below that will make sure it doesn't interfer with your existing config:

```lua
-- DO NOT change the paths and don't remove the colorscheme
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", lazypath, })
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
  "folke/tokyonight.nvim",
  "hrsh7th/nvim-cmp",
  -- add any other plugins here
}
require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

vim.cmd.colorscheme("tokyonight")
-- add anything else here
```

zaqxsw-dev:
@folke @hrsh7th
1. you can take my config from issue https://github.com/hrsh7th/nvim-cmp/issues/1608
2. install rust ```curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh```
3. install rust-analyzer ```rustup component add rust-analyzer``` or for macos ```brew install rust-analyzer```
4. create new project ```cargo new test-rs```
5. start edit src/main.rs

Replace code in main.rs to this code
```rust
pub fn remove_char(s: &str) -> String {
    (&(String::from(s))[0..s.len()]).to_string()
}
fn main() {
    println!("Hello, world!");
    remove_char(START_TYPING_HERE);
}
```
Start typing args in remove_char 
... get input delay.

zaqxsw-dev:
@folke 
```lua
require('packer').startup(function(use)
use { "wbthomason/packer.nvim" }
use {'hrsh7th/nvim-cmp'}
use {'hrsh7th/cmp-nvim-lsp'}
use {'hrsh7th/cmp-buffer'}
use {'hrsh7th/cmp-path'}
use {'hrsh7th/cmp-cmdline'}
use {'ray-x/lsp_signature.nvim'}
use {'neovim/nvim-lspconfig'}
use {'simrat39/rust-tools.nvim'}
use {'saadparwaiz1/cmp_luasnip'}
end)

local cmp = require'cmp'

cmp.setup({
  mapping = {
    ['<Tab>'] = function(fallback)
      if cmp.visible() then
        cmp.select_next_item()
      else
        fallback()
      end
    end,
    ['<C-f>'] = cmp.mapping(cmp.mapping.scroll_docs(-4), { 'i', 'c' }),
    ['<C-c>'] = cmp.mapping(cmp.mapping.scroll_docs(4), { 'i', 'c' }),
    ['<CR>'] = cmp.mapping.confirm({ select = true }), -- Accept currently selected item. Set `select` to `false` to only confirm explicitly selected items.
  },
  snippet = {
    expand = function(args)
      require('luasnip').lsp_expand(args.body)
    end,
  },
  sources = cmp.config.sources(
    {
      { name = 'nvim_lsp' },
    },
    {
      { name = 'buffer' },
    }
  )
})

cmp.setup.cmdline({ '/', '?' }, {
  sources = {
    { name = 'buffer' }
  }
})

cmp.setup.cmdline(':', {
    { name = 'path' }
  }, {
    { name = 'cmdline' }
  })
})

local rt = require("rust-tools")

rt.setup({
  server = {
    capabilities = capabilities,
    on_attach = function(_, bufnr)
      vim.keymap.set("n", "<Leader>s", rt.hover_actions.hover_actions, { buffer = bufnr })
      vim.keymap.set("n", "<Leader>a", rt.code_action_group.code_action_group, { buffer = bufnr })
    end,
  },
})

local cfg = {
  floating_window = false,
  hint_prefix = "🎇 ",
}
require "lsp_signature".setup(cfg)
```

folke:
Just created a proper fix in #1611
Would be great if you could confirm.

Issue #273, opened 2021-10-05
How to limit the maximum number of automatic reminder bars


xuliefeng:
How to limit the maximum number of automatic reminder bars. There are too many columns currently displayed, and only the top columns can be seen. So is it possible to limit the total number displayed

hrsh7th:
You can use `set pumheight=10`

PR #1611, opened 2023-06-08
fix(async): properly trigger callbacks when canceling and fix delays in `throttle.sync`

folke:
Fixes #1610 #1608 #1610

### DIFF
--- a/lua/cmp/utils/async.lua
+++ b/lua/cmp/utils/async.lua
@@ -288,7 +288,7 @@ end
 
 -- This will yield when called from a coroutine
 function async.yield(...)
-  if not coroutine.isyieldable() then
+  if coroutine.running() == nil then
     error('Trying to yield from a non-yieldable context')
     return ...
   end
--- a/lua/cmp/utils/async.lua
+++ b/lua/cmp/utils/async.lua
@@ -32,10 +32,14 @@ async.throttle = function(fn, timeout)
   local timer = assert(vim.loop.new_timer())
   local _async = nil ---@type Async?
   timers[#timers + 1] = timer
-  return setmetatable({
+  local throttle
+  throttle = setmetatable({
     running = false,
     timeout = timeout,
     sync = function(self, timeout_)
+      if not self.running then
+        return
+      end
       vim.wait(timeout_ or 1000, function()
         return not self.running
       end)
@@ -44,6 +48,8 @@ async.throttle = function(fn, timeout)
       if reset_time ~= false then
         time = nil
       end
+      -- can't use self here unfortunately
+      throttle.running = false
       timer:stop()
       if _async then
         _async:cancel()
@@ -57,9 +63,8 @@ async.throttle = function(fn, timeout)
       if time == nil then
         time = vim.loop.now()
       end
-
-      self.running = true
       self.stop(false)
+      self.running = true
       timer:start(math.max(1, self.timeout - (vim.loop.now() - time)), 0, function()
         vim.schedule(function()
           time = nil
@@ -80,6 +85,7 @@ async.throttle = function(fn, timeout)
       end)
     end,
   })
+  return throttle
 end
 
 ---Control async tasks.
@@ -217,12 +223,18 @@ end
 ---@param result? any
 ---@param error? string
 function Async:_done(result, error)
-  self.running = false
-  self.result = result
-  self.error = error
+  if self.running then
+    self.running = false
+    self.result = result
+    self.error = error
+  end
   for _, callback in ipairs(self.callbacks) do
     callback(result, error)
   end
+  -- only run each callback once.
+  -- _done can possibly be called multiple times.
+  -- so we need to clear callbacks after executing them.
+  self.callbacks = {}
 end
 
 function Async:_step()
@@ -237,7 +249,7 @@ function Async:_step()
 end
 
 function Async:cancel()
-  self.running = false
+  self:_done(nil, 'abort')
 end
 
 ---@param cb AsyncCallback
--- a/lua/cmp/utils/async.lua
+++ b/lua/cmp/utils/async.lua
@@ -42,7 +42,7 @@ async.throttle = function(fn, timeout)
       end
       vim.wait(timeout_ or 1000, function()
         return not self.running
-      end)
+      end, 10)
     end,
     stop = function(reset_time)
       if reset_time ~= false then
@@ -73,6 +73,7 @@ async.throttle = function(fn, timeout)
             ---@cast ret Async
             _async = ret
             _async:await(function(_, error)
+              _async = nil
               self.running = false
               if error and error ~= 'abort' then
                 vim.notify(error, vim.log.levels.ERROR)
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -383,12 +383,11 @@ entry.match = function(self, input, matching_config)
     score, matches = matcher.match(input, filter_text, option)
 
     -- Support the language server that doesn't respect VSCode's behaviors.
-    local prefix = ''
     if score == 0 then
       if self:get_completion_item().textEdit and not misc.empty(self:get_completion_item().textEdit.newText) then
         local diff = self.source_offset - self:get_offset()
         if diff > 0 then
-          prefix = string.sub(self.context.cursor_line, self:get_offset(), self:get_offset() + diff)
+          local prefix = string.sub(self.context.cursor_line, self:get_offset(), self:get_offset() + diff)
           local accept = nil
           accept = accept or string.match(prefix, '^[^%a]+$')
           accept = accept or string.find(self:get_completion_item().textEdit.newText, prefix, 1, true)


folke:
draft for now, since I'm not able to reproduce the ray lsp signature issue for now.

lua/cmp/utils/async.lua
```diff
@@ -58,8 +61,8 @@ async.throttle = function(fn, timeout)
         time = vim.loop.now()
```
hrsh7th, line 61: IMO, we should add `this.running = false` to `stop`.

lua/cmp/utils/async.lua
```diff
@@ -58,8 +61,8 @@ async.throttle = function(fn, timeout)
...
       end
 
-      self.running = true
       self.stop(false)
+      self.running = true
```
mgnsk, line 65: When I changed to `self.running = false`, the issue is solved and the stutter is gone. (context: https://github.com/hrsh7th/nvim-cmp/pull/1610#issuecomment-1583286858)

lua/cmp/utils/async.lua
folke: I agree. Just fixed that.

lua/cmp/utils/async.lua
folke: That's not a real fix unfortunately.
I did however fix another issue in the code just now. Can you try again using this PR?

folke:
@hrsh7th the way the throttle works right now, is that it will wait to restart based on how much time there's left till next execution. This also means that `self.running` will be `true` until the next run is scheduled and executed.

`core.filter` is always an async throttle.

`cmp.sync` is used in a lot of places with a `callback`. It would be much better to use `async:await(cb)` instead, so we don't lock the main thread with `vim.wait`.

But the problem remains that the throttle might just be waiting till next execution.

Could still be solved by `throttle.await` saving callbacks to run once `_async` starts and then `:await` on that one.

`async.throttle` is currently only used in two places.

Maybe we should rework the whole `async.throttle` function and create one specifically for async functions? Specifically for `filter` in this case? `async.throttle` can now work with both regular and async functions.

lua/cmp/utils/async.lua
mgnsk: Can't get it to work anymore:
```
Error executing vim.schedule lua callback: .../magnus/.local/share/nvim/lazy/nvim-cmp/lua/cmp/core.lua:294: attempt to index field 'filter' (a nil value)
stack traceback:
        .../magnus/.local/share/nvim/lazy/nvim-cmp/lua/cmp/core.lua:294: in function 'complete'
        .../magnus/.local/share/nvim/lazy/nvim-cmp/lua/cmp/core.lua:170: in function 'autoindent'
        .../magnus/.local/share/nvim/lazy/nvim-cmp/lua/cmp/core.lua:162: in function 'on_change'
        .../magnus/.local/share/nvim/lazy/nvim-cmp/lua/cmp/init.lua:313: in function 'callback'
        .../.local/share/nvim/lazy/nvim-cmp/lua/cmp/utils/async.lua:164: in function <.../.local/share/nvim/lazy/nvim-cmp/lua/cmp/utils/async.lua:162>
```

folke: just fixed that, but something is very broken apart from that with my PR. Will check what's wrong here

folke: Should be fixed now. I wasnt using this branch myself. I thought I still had lazy configured to use the dev version :)

Can you test with my latest changes?

mgnsk: I tried https://github.com/hrsh7th/nvim-cmp/pull/1611/commits/0d8de773e08aca4e3be665a0ee8895127a25e36f but still stutters. Using the correct config, deleted `lazy-lock.json` and `nvim-cmp` plugin itself and ran `Lazy sync`.
```lua     {
        "folke/nvim-cmp",
        branch = "filter_sync",
```

lua/cmp/utils/async.lua
```diff
@@ -32,18 +32,24 @@ async.throttle = function(fn, timeout)
...
     running = false,
     timeout = timeout,
     sync = function(self, timeout_)
+      if not self.running then
+        return
+      end
       vim.wait(timeout_ or 1000, function()
         return not self.running
-      end)
+      end, 10)
```
folke, line 45: This is the real fix for the delay some users experienced.

Reason is that previsouly the code was not async, so filter would not be running in most cases.
With async, we are hitting the wait in `cmp.sync`.

Default interval is `200ms` for `vim.wait` which was causing the delays.

lua/cmp/utils/async.lua
folke: Found the issue and should be fixed now. Would be great if you can confirm

lua/cmp/utils/async.lua
zaqxsw-dev: thank you, i will check it now

folke: @hrsh7th instead of `10`, it could be set to an even lower value, but I personally don't see the delays anymore in the repro.

mgnsk: It's still pretty bad, I tried this:
```lua
cmp.visible = cmp.sync(function()
  return cmp.core.view:visible() or vim.fn.pumvisible() == 1
end)
```
to
```lua
cmp.visible = function()
  return cmp.core.view:visible() or vim.fn.pumvisible() == 1
end
```

and the delay is gone. I have no idea about the architecture of `nvim-cmp` and why a visibility check would require so much code.

folke: @mgnsk the only repro that was provided to me is [here](https://github.com/hrsh7th/nvim-cmp/pull/1610#issuecomment-1584280889)

Those delays are fixed by this PR and multiple people confirmed it.

Are you 100% you're on the latest nvim-cmp version?

If the issue persists, then I really need a repro to check.

mgnsk: I'll look into it more closely in the evening. I'm sure I saw this PR title when I did `Lazy sync`.

eeeXun:
I'm still get laggy after this patch.

```
:version
NVIM v0.9.1
Build type: Release
LuaJIT 2.1.0-beta3

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/share/nvim"

Run :checkhealth for more info
```

<details>
<summary>/tmp/mini.lua</summary>

```lua
vim.cmd([[set runtimepath=$VIMRUNTIME]])
vim.cmd([[set packpath=/tmp/cmp-min/site]])
local package_root = "/tmp/cmp-min/site/pack"
local install_path = package_root .. "/packer/start/packer.nvim"
local function load_plugins()
    require("packer").startup({
        {
            "wbthomason/packer.nvim",
            {
                "williamboman/mason.nvim",
                config = function()
                    require("mason").setup({
                        install_root_dir = "/tmp/cmp-min/data",
                    })
                end,
            },
            {
                "WhoIsSethDaniel/mason-tool-installer.nvim",
                config = function()
                    require("mason-tool-installer").setup({
                        ensure_installed = {
                            "typescript-language-server",
                        },
                    })
                end,
            },
            {
                "neovim/nvim-lspconfig",
                config = function()
                    local capabilities = require("cmp_nvim_lsp").default_capabilities()
                    require("lspconfig")["tsserver"].setup({
                        capabilities = capabilities,
                    })
                end,
            },
            {
                "ray-x/lsp_signature.nvim",
                config = function()
                    require("lsp_signature").setup({
                        bind = true,
                        toggle_key = "<C-s>",
                        toggle_key_flip_floatwin_setting = true,
                    })
                end,
            },
            {
                "hrsh7th/nvim-cmp",
                requires = { "hrsh7th/cmp-nvim-lsp" },
                config = function()
                    local cmp = require("cmp")
                    cmp.setup({
                        mapping = {
                            ["<CR>"] = cmp.mapping.confirm({ select = true }),
                        },
                            { name = "nvim_lsp" },
                        }),
                    })
                end,
            },
        },
        config = {
            package_root = package_root,
            compile_path = install_path .. "/plugin/packer_compiled.lua",
            display = { non_interactive = true },
        },
    })
end
if vim.fn.isdirectory(install_path) == 0 then
    vim.fn.system({ "git", "clone", "--depth=1", "https://github.com/wbthomason/packer.nvim", install_path })
end
load_plugins()
require("packer").sync()
```
</details>

Run with `nvim -u /tmp/mini.lua`. As run it first time, we should quit nvim after plugins are installed. And open it again `nvim -u /tmp/mini.lua` to let `mason-tool-installer.nvim` auto install `typescript-language-server`. Then edit a js file.

https://github.com/hrsh7th/nvim-cmp/assets/58657914/9dd856ad-d429-48a7-b777-f1fd5a2a637f

As you can see. This doesn't `always` happen. But quite `frequently` for me.

BTW. I've checked that removing `lsp_signature.nvim`,  input will not get stuck.

mgnsk:
> I'm still get laggy after this patch.

@eeeXun See https://github.com/hrsh7th/nvim-cmp/issues/1613.

Your video is even more clear because it shows which keys were pressed before the lag happened.

basilgood:
Same here, it's better but still get laggy:  [#273](https://github.com/ray-x/lsp_signature.nvim/issues/273#issuecomment-1585523599)

Issue #1433, opened 2023-02-11
Performance problems occur when a single line is huge


fengwk:
Phenomenon: When I paste a huge json into nvim, nvim will get stuck in the input near the end of the line in insert mode.

Cause: I made some positioning and found the following functions in `core.lua`. The performance problem lies in the regex matching of `pattern.matchstr('[^[:blank:]]\\+$', cursor_before_line)`.

```
---Check autoindent
---@param trigger_event cmp.TriggerEvent
---@param callback function
core.autoindent = function(self, trigger_event, callback)
  if trigger_event ~= types.cmp.TriggerEvent.TextChanged then
    return callback()
  end
  if not api.is_insert_mode() then
    return callback()
  end

  -- Check prefix
  local cursor_before_line = api.get_cursor_before_line()
  local prefix = pattern.matchstr('[^[:blank:]]\\+$', cursor_before_line) or ''
  if #prefix == 0 then
    return callback()
  end

  -- Reset current completion if indentkeys matched.
  for _, key in ipairs(vim.split(vim.bo.indentkeys, ',')) do
    if vim.tbl_contains({ '=' .. prefix, '0=' .. prefix }, key) then
      self:reset()
      self:set_context(context.empty())
      break
    end
  end

  callback()
end
```

jalemann:
Hey there this still seems to be a problem for me, opening a single line json of around 1MB file size and starting to enter a nvim command, nvim freezes for around 3 minutes and uses a few GB of memory.

Workaround: I can use ctrl-c to cancel keyboard interrupt the cmp regex and continue as usual.

Still wanted to report, I am not sure if that is how it is supposed to be now and if it froze even longer before this fix or if there is something wrong still...
I am using nvim-cmp up to commit 0b751f6beef40fd47375eaf53d3057e0bfa317e4.

dyfrgi:
@jalemann Take a look at #1841. I found that for similar files, I had to disable the locality sorter to get good performance.

PR #1615, opened 2023-06-10
fix: huge single line performance (#1433)

fengwk:
Solve the performance issue of #1433.

I wrote a function with the same semantics as the regular expression to replace it. The function searches for a non-empty character suffix starting from the end of the line.

Since the string is huge, it still appears slow, but it can avoid freezing.


hrsh7th:
thank you!

PR #1616, opened 2023-06-10
fix: allow ghost_text type to accept values of true

tj-moody:
Fix type of ghost_text to not create a warning when set to true
Screenshot 2023-06-10 at 11 44 59 AM

The existing type appears to be forcing the user to supply a table containing a hl_group for ghost text by specifying a value of false otherwise,
Screenshot 2023-06-10 at 11 54 16 AM
and yet the ghost text implementation clearly allows a value of ghost_text other than a table with hl_group by defaulting to 'Comment'
Screenshot 2023-06-10 at 11 55 28 AM


hrsh7th:
thank you

PR #1624, opened 2023-06-17
fix type of the config.autocomplete to accept false

kyoh86:
I have fixed it to match the following as described in the help

> completion.autocomplete
>   `cmp.TriggerEvent[] | false`


hrsh7th:
Thank you!

PR #1626, opened 2023-06-18
Expand docs for select_next_item select_prev_item

Sleepful:
this took me a while to figure out, see here:

https://github.com/hrsh7th/nvim-cmp/discussions/1625


hrsh7th:
Thank you!

PR #1646, opened 2023-07-05
Add toggle documentation functionality

### LINKED ISSUE #674: [Feature Request] Toggle Document
I would like to have nvim-cmp always select the first item, but it opens the doc on hover. I would like to only trigger the doc when I press a key combination, so I can have the first item always selected and not having to press the down key two times to navigate to the second item.

MariaSolOs:
Fixes #674 

I haven't documented yet how the setting works since I first want to get a thumbs up from @hrsh7th, but the idea is:

<s>- If `completion.doc_visibility` is `'OnEntryChange'`, then the documentation window will appear when the currently selected entry changes.
- If set to `'Toggle'`, then it is assumed that the user will create the corresponding mapping to open/close the window.
- If set to `'Disabled'`, then the window will never open.

This might be confusing with the current way to disable the documentation window via `window.documentation`, so if this goes in we might want to deprecate that setting.</s>

### DIFF
--- a/lua/cmp/config/mapping.lua
+++ b/lua/cmp/config/mapping.lua
@@ -178,6 +178,15 @@ mapping.scroll_docs = function(delta)
   end
 end
 
+-- Toggle the documentation window.
+mapping.toggle_docs = function()
+  return function(fallback)
+    if not require('cmp').toggle_docs() then
+      fallback()
+    end
+  end
+end
+
 ---Select next completion item.
 mapping.select_next_item = function(option)
   return function(fallback)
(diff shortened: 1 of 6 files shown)


MariaSolOs:
Found a couple of issues. Will publish again soon.

hrsh7th:
I'll review it. Thank you.

lua/cmp/config/mapping.lua
```diff
@@ -178,6 +178,15 @@ mapping.scroll_docs = function(delta)
...
 end
 
+-- Toggle the documentation window.
+mapping.toggle_docs = function()
+  return function(fallback)
+    if not require('cmp').toggle_docs() then
+      fallback()
+    end
+  end
+end
```
hrsh7th, line 188: I think it should be `open/close` instead of `toggle`.

MariaSolOs:
Ahhh for some reason GitHub isn't showing all my commits. I'll open another PR.

lua/cmp/config/mapping.lua
MariaSolOs: You mean I shouldn't create a new toggle function?
Note that I changed the implementation a bit in #1647, since I realized that this first attempt wouldn't properly update the documentation window when navigating through the list.

PR #1647, opened 2023-07-05
feat: add toggle_doc functionality

MariaSolOs:
Refer to #1646 

I'm very sorry for the trouble, for some reason GitHub stopped showing all my commits (??).

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -41,6 +41,7 @@ return function()
       completeopt = 'menu,menuone,noselect',
       keyword_pattern = [[\%(-\?\d\+\%(\.\d\+\)\?\|\h\w*\%(-\w*\)*\)]],
       keyword_length = 1,
+      docs_initially_visible = true,
     },
 
     formatting = {
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -95,6 +95,9 @@ return function()
         name = 'custom',
         selection_order = 'top_down',
       },
+      docs = {
+        auto_open = true,
+      }
     },
 
     window = {
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -41,7 +41,6 @@ return function()
       completeopt = 'menu,menuone,noselect',
       keyword_pattern = [[\%(-\?\d\+\%(\.\d\+\)\?\|\h\w*\%(-\w*\)*\)]],
       keyword_length = 1,
-      docs_initially_visible = true,
     },
 
     formatting = {
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -10,7 +10,6 @@ local ghost_text_view = require('cmp.view.ghost_text_view')
 
 ---@class cmp.View
 ---@field public event cmp.Event
----@field public docs_visible boolean
 ---@field private resolve_dedup cmp.AsyncDedup
 ---@field private native_entries_view cmp.NativeEntriesView
 ---@field private custom_entries_view cmp.CustomEntriesView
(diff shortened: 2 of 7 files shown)


hrsh7th:
The implementation is good but I think the API should be `cmp.visible_docs()`, `cmp.open_docs()` and `cmp.close_docs()`.

What do you think?

lua/cmp/config/default.lua
```diff
@@ -41,6 +41,7 @@ return function()
       completeopt = 'menu,menuone,noselect',
       keyword_pattern = [[\%(-\?\d\+\%(\.\d\+\)\?\|\h\w*\%(-\w*\)*\)]],
       keyword_length = 1,
+      docs_initially_visible = true,
```
hrsh7th, line 44: Hm... I'll consider the better place for this option.

MariaSolOs:
> The implementation is good but I think the API should be `cmp.visible_docs()`, `cmp.open_docs()` and `cmp.close_docs()`.
> 
> What do you think?

Sounds reasonable to me!

lua/cmp/config/default.lua
MariaSolOs: Yeah I was unsure about it too. IMO it doesn't make sense to put it in `WindowConfig` since those only involve UI settings, and all other categories don't feel right either...

MariaSolOs: We can also just get rid of this option and let users configure if docs should be initially visible or not.

MariaSolOs:
@hrsh7th lmk if this looks good to you. I can add a couple of examples and update the documentation to explain how these work :)

lua/cmp/view.lua
```diff
@@ -239,7 +259,9 @@ view.on_entry_change = async.throttle(function(self)
       if not self:visible() then
         return
       end
-      self.docs_view:open(e, self:_get_entries_view():info())
+      if self.docs_view:visible() then
```
hrsh7th, line 262: Sorry. I didn't understand why this was needed. Could you explain it?

hrsh7th:
Almost looks good.

lua/cmp/view.lua
MariaSolOs: This check is needed in order to correctly update the documentation window when the selected entry changes.

- If the documentation window is currently closed, then it won't be opened when the entry changes.
- If it's open, then we ensure that its content reflects the new entry.

MariaSolOs: Hmm the issue with this is that the docs will initially be hidden. And so I think we do need the `docs_initially_visible`, or do you have a better idea of how to address this issue?

MariaSolOs: @hrsh7th what are your thoughts on this :) ?

hrsh7th:
sorry for the delay. I've been busy with work...😭

I'll push a new commit for this PR. Please wait.

MariaSolOs:
> sorry for the delay. I've been busy with work...😭
> 
> I'll push a new commit for this PR. Please wait.

No worries! Take your time, I’m just trying to help :)

MariaSolOs:
Any update on this? :)

hrsh7th:
I pushed a fix commit. Can you please confirm?

MariaSolOs:
> I pushed a fix commit. Can you please confirm?

I just tried it and it works great :) Thank you so much for taking the time to improve and document the feature!

astier:
I use

``` lua
local cmp = require('cmp')
cmp.setup({
  mapping = cmp.mapping.preset.insert({
    ['K'] = cmp.mapping(function(fallback)
      if cmp.visible_docs() then
        cmp.close_docs()
      elseif cmp.visible() then
        cmp.open_docs()
      else
        fallback()
      end
    end),
  })
})
```

MariaSolOs:
@beh-10257 read the docs: https://github.com/hrsh7th/nvim-cmp/blob/a110e12d0b58eefcf5b771f533fc2cf3050680ac/doc/cmp.txt#L970

PR #1650, opened 2023-07-09
perf: fix `nvim_replace_termcodes` being called on every `CursorMoved`

sandersantema:
Previously `vim.api.nvim_replace_termcodes('<C-c>', true, true, true)` was called every time the `CursorMoved` autocommand event fired. Even though afaik there is no reason to do so as the termcode stays the same. I discovered this by profiling with [profile.nvim](https://github.com/stevearc/profile.nvim) while trying to fix another issue (unrelated to cmp).

Feel free to change the variable name.

### DIFF
--- a/plugin/cmp.lua
+++ b/plugin/cmp.lua
@@ -41,8 +41,9 @@ end)
 autocmd.emit('ColorScheme')
 
 if vim.on_key then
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you.

PR #1675, opened 2023-08-06
perf(core): simplify and improve `find_line_suffix()`

tmillr:
Call `string.match()` once instead of calling `string.sub()` and `string.find()` multiple times in a loop.


hrsh7th:
Thank you!

Issue #1113, opened 2022-08-03
Buffer is not 'modifiable'


bugsbugsbux:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = {
    ['<CR>'] = cmp.mapping.confirm({ select = true })
  },

  sources = {
    { name = "nvim_lsp" },
    { name = "buffer" },
  },
}

cmp.setup.cmdline(':', {
    mapping = cmp.mapping.preset.cmdline({}),
    sources = cmp.config.sources({
        {name = 'nvim_lsp'},
        {name = 'buffer'},
    }),
})
vim.cmd[[autocmd FileType markdown nnoremap <buffer> <F5> <cmd>pedit +set\ nomodifiable <cfile><cr>]]
EOF

lua << EOF
local capabilities = require('cmp_nvim_lsp').update_capabilities(vim.lsp.protocol.make_client_capabilities())

require'lspconfig'.cssls.setup {
  capabilities = capabilities,
}
EOF
```

### Description

[...]

PR #1677, opened 2023-08-14
Fix: Buffer is not 'modifiable'

GyouzaRu:
### Description
When I try to open a file with `-M` option. 
**command:** `nvim -M <filename>`.
The completion of commandline will be break.

### Root cause
I found that `nvim_buf_set_lines` cannot work when the buffer is `unmodifiable`.

### Solution
Add judgement before 'nvim_buf_set_lines': if the buffer is `unmodifiable`, make it `modifiable` temporarily.

### Related Issue
I think #1113 is relative to this problem.

### DIFF
--- a/lua/cmp/view/custom_entries_view.lua
+++ b/lua/cmp/view/custom_entries_view.lua
@@ -140,7 +140,13 @@ custom_entries_view.open = function(self, offset, entries)
       end
     end
   end
-  vim.api.nvim_buf_set_lines(entries_buf, 0, -1, false, lines)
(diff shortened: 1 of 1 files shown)


max397574:
couldn't you just use `vim.bo`?
would be much simpler and more readable code imo

GyouzaRu:
> couldn't you just use `vim.bo`? would be much simpler and more readable code imo

Thanks for your advice, that's very helpful.

hrsh7th:
Thank you.

PR #1687, opened 2023-08-24
docs: Add comments and type annotations for sorting comparators

wookayin:
Also added type annotation `cmp.Comparator` for sorting.comparators


hrsh7th:
thank you.

Issue #748, opened 2022-01-19
Use `FloatBorder` instead of `NormalFloat` for highlighting borders


SimonMandlik:
As it is, it is only possible to change the color of both normal text and border at once, e.g

<img width="581" alt="image" src="https://user-images.githubusercontent.com/16707112/150184908-58debd59-3170-4f9f-9ebe-f3a5916d817b.png">

SimonMandlik:
Don't really know why, but this bit of code in the config helps

```
    documentation = {
        border = { "╭", "─", "╮", "│", "╯", "─", "╰", "│" },
        winhighlight = 'NormalFloat:NormalFloat,FloatBorder:FloatBorder',
    },
```

esn89:
@simonmandlik 
I have the same question as you.

Can I see your code or nvim configuration where you set the hover/Documentation window fg and bg?

simonmandlik:
Hi, I am using the following hg definitions

```
highlight CmpItemAbbr guibg=NONE
highlight! link CmpItemMenu LineNr
highlight! link CmpItemAbbrMatch Bold
highlight! link CmpItemAbbrMatchFuzzy Bold
highlight CmpItemAbbrDeprecated gui=strikethrough

highlight! link CmpItemKindConstant TSConstant
highlight! link CmpItemKindConstructor TSConstructor
highlight! link CmpItemKindField TSField
highlight! link CmpItemKindFunction TSFunction
highlight! link CmpItemKindKeyword TSKeyword
highlight! link CmpItemKindMethod TSMethod
highlight! link CmpItemKindOperator TSOperator
highlight! link CmpItemKindProperty TSProperty
highlight! link CmpItemKindText TSText
highlight! link CmpItemKindVariable TSVariable

highlight! link CmpItemKindSnippet TSString

highlight! link CmpItemKindEnum TSVariable
highlight! link CmpItemKindEnumMember TSField

highlight! link CmpItemKindModule TSNamespace
highlight! link CmpItemKindUnit TSNamespace

highlight! link CmpItemKindClass TSType
highlight! link CmpItemKindStruct TSType
highlight! link CmpItemKindInterface TSType

highlight! link CmpItemKindValue TSLiteral

highlight! link CmpItemKindFile TSNote
highlight! link CmpItemKindFolder TSNote

highlight! link CmpItemKindEvent TSWarning

highlight! link CmpItemKindColor TSAnnotation

highlight! link CmpItemKindReference TSLabel

[...]

PR #1689, opened 2023-08-26
feat: match default border color to standard neovim float border hl

ribru17:
This commit differentiates the border color from the inner content color of the nvim-cmp float. The `FloatBorder` group links to `VertSplit` by default and is usually set by the colorscheme, so this commit makes nvim-cmp match with other float window colors. 

Say this is our usual LSP float window:
lspfloatexample

Previously our cmp window looks like this:
cmpbefore

Now it will match the LSP window (and other floating windows):
cmpafter

This was discussed in #748 but was closed without any change.

NOTE: This commit changes the default for `config.window.bordered`, but the borderless default remains the same since there is no border there anyway. If one was added the border would look as it does in the old way but at that point the user would probably be configuring everything manually anyway.

### DIFF
--- a/lua/cmp/config/window.lua
+++ b/lua/cmp/config/window.lua
@@ -4,7 +4,7 @@ window.bordered = function(opts)
   opts = opts or {}
   return {
     border = opts.border or 'rounded',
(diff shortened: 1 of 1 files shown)


hrsh7th:
Thank you!

PR #1694, opened 2023-09-05
Mark ConfigSchema fields as optional to prevent LSP warnings

HNegus:
The Lua Language Server will show a warning when calling `cmp.setup` because the table fields are not marked as optional.


wookayin:
Then all the internal codes that depend on `cmp.ConfigSchema` should do (unnecessary) nil check, right? I think a better solution might be have a **different** type than `cmp.ConfigSchema` that is used for `cmp.setup` only.

HNegus:
I believe the type is used in just a few places, so wouldn't the impact be very limited?

hrsh7th:
Sorry. I'd merged another PR. Thank you.

wookayin:
Fixed in #1723

PR #1723, opened 2023-10-09
Mark non-required setup fields as optional

komidore64:
I have been getting "Missing required fields" messages from lua-language-server (`> 3.6.23`) for table values handed to `require('cmp').setup()`.

missing-required-fields-diagnostics

This is a naive attempt to fix those missing required fields messages. I'm still fairly new to Lua, but happy to learn! 

Please let me know if I've missed any others that are optional so I can get those marked too.

Once applying this patch to a local checkout of `nvim-cmp`, I don't get missing field diagnostic messages anymore.

no-diagnostics

Thanks for considering my patch!

### DIFF
--- a/lua/cmp/types/cmp.lua
+++ b/lua/cmp/types/cmp.lua
@@ -83,21 +83,21 @@ cmp.ItemField = {
 ---@alias cmp.Mapping cmp.MappingFunction | cmp.MappingClass
 
 ---@class cmp.ConfigSchema
----@field private revision integer
----@field public enabled boolean | fun(): boolean
----@field public performance cmp.PerformanceConfig
----@field public preselect cmp.PreselectMode
----@field public completion cmp.CompletionConfig
----@field public window cmp.WindowConfig|nil
----@field public confirmation cmp.ConfirmationConfig
----@field public matching cmp.MatchingConfig
----@field public sorting cmp.SortingConfig
----@field public formatting cmp.FormattingConfig
----@field public snippet cmp.SnippetConfig
----@field public mapping table<string, cmp.Mapping>
----@field public sources cmp.SourceConfig[]
----@field public view cmp.ViewConfig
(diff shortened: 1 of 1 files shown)


lttb:
I have the same issue, it'd be great to see it merged! thank you

tjdevries:
@hrsh7th this does seem like a pretty good change IMO, not sure if there is a more elegant way to solve this but it's nice to note that the user does not HAVE to provide each of the fields

hrsh7th:
Thank you!

lua/cmp/types/cmp.lua
```diff
@@ -83,21 +83,21 @@ cmp.ItemField = {
...
+---@field public sorting? cmp.SortingConfig
+---@field public formatting? cmp.FormattingConfig
+---@field public snippet? cmp.SnippetConfig
+---@field public mapping? table<string, cmp.Mapping>
+---@field public sources? cmp.SourceConfig[]
+---@field public view? cmp.ViewConfig
+---@field public experimental? cmp.ExperimentalConfig
 
 ---@class cmp.PerformanceConfig
 ---@field public debounce integer
```
liskin, line 103: Any reason not to do the same thing here?

lua/cmp/types/cmp.lua
```diff
@@ -108,24 +108,24 @@ cmp.ItemField = {
...
+---@field public border? string|string[]
+---@field public winhighlight? string
+---@field public zindex? integer|nil
+---@field public max_width? integer|nil
+---@field public max_height? integer|nil
+---@field public scrolloff? integer|nil
+---@field public scrollbar? boolean|true
 
 ---@class cmp.ConfirmationConfig
 ---@field public default_behavior cmp.ConfirmBehavior
```
liskin, line 131: Likewise. My config only sets `get_commit_characters` and lua_ls complains…

Issue #1705, opened 2023-09-16
Astro LSP completions sometimes write over incorrect ranges


Trildar:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```Lua
local lazypath = vim.fn.stdpath("data") .. "/lazy/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  -- bootstrap lazy.nvim
  -- stylua: ignore
  vim.fn.system({ "git", "clone", "--filter=blob:none", "https://github.com/folke/lazy.nvim.git", "--branch=stable", lazypath })
end
vim.opt.rtp:prepend(vim.env.LAZY or lazypath)

require("lazy").setup({
	spec = {
		{
			"williamboman/mason.nvim",
			cmd = "Mason",
			keys = { { "<leader>cm", "<cmd>Mason<cr>", desc = "Mason" } },
			build = ":MasonUpdate",
			opts = {
				ensure_installed = {
					"astro-language-server",
				},
			},
			---@param opts MasonSettings | {ensure_installed: string[]}
			config = function(_, opts)
				require("mason").setup(opts)
				local mr = require("mason-registry")
				local function ensure_installed()
					for _, tool in ipairs(opts.ensure_installed) do
						local p = mr.get_package(tool)
						if not p:is_installed() then
							p:install()
						end
					end
				end
				if mr.refresh then
					mr.refresh(ensure_installed)
				else
					ensure_installed()
				end
			end,
		},
		"hrsh7th/cmp-nvim-lsp",
		{
			"hrsh7th/nvim-cmp",
			version = false, -- last release is way too old
			event = "InsertEnter",
			dependencies = {
				"hrsh7th/cmp-nvim-lsp",
			},
			opts = function()
				vim.api.nvim_set_hl(0, "CmpGhostText", { link = "Comment", default = true })
				local cmp = require("cmp")
				local defaults = require("cmp.config.default")()
				return {
					completion = {
						completeopt = "menu,menuone,noinsert",
					},
					mapping = cmp.mapping.preset.insert({
						["<C-n>"] = cmp.mapping.select_next_item({ behavior = cmp.SelectBehavior.Insert }),
[...]

PR #1725, opened 2023-10-12
Favor existing values in completion_item over resovled_completion_item

Trildar:
Fix/workaround for #1705

"Workaround" because I think ultimately the issue is in the Astro language server or an upstream language server returning wrong ranges for `completionItem/resolve` requests.

Excuse me if there is a reason this behaviour is undesirable. Since the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion) says that properties provided in `textDocument/completion` "must not be changed during resolve" and VS Code presumably (I don't know where its code for handling completions is) behaves like this, as it's not affected by the above issue, I kind of assume this would be fine.

Also, just out of curiosity, is there a reason this doesn't use `vim.tbl_extend()`?

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -423,7 +423,7 @@ entry.get_completion_item = function(self)
     if self.resolved_completion_item then
       local completion_item = misc.copy(self.completion_item)
       for k, v in pairs(self.resolved_completion_item) do
-        completion_item[k] = v or completion_item[k]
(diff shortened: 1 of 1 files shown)


phaberest:
I can confirm this fixes that wrong behaviour, thank you @Trildar (at least I was able to fix my own environment)

hrsh7th:
I didn't know it was specified in the LSP specification! thank you.

PR #1727, opened 2023-10-15
feat: add option for view to follow cursor

### LINKED ISSUE #1660: [Question] Completition window popup following the cursor
It's possible to make the competition window popup, following the cursor while type?

mcauley-penney:
~~Creates an option to allow the custom entries
view to follow the user's cursor as they type.~~

This PR adds an option to allow the view, both custom and native, to follow the user's cursor as they type.

To enable, set
```lua
require("cmp").setup({
  view = {
    entries = {
      follow_cursor = true
    }
  }
})
```

Original custom view source at https://github.com/lvimuser/nvim-cmp/commit/7569056388417d887baf2e959e18a767bcfe84f1

Closes #1660

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -679,6 +679,13 @@ view.docs.auto_open~
 
   Specify whether to show the docs_view when selecting an item.
 
+                                                 *cmp-config.view.follow_cursor*
+view.follow_cursor*
+  `boolean`
+
+  Specify whether the pmenu should follow the current position of the cursor
+  as the user types. `false` by default.
+
                            *cmp-config.window.{completion,documentation}.border*
 window.{completion,documentation}.border~
   `string | string[] | nil`
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -94,6 +94,7 @@ return function()
       entries = {
         name = 'custom',
         selection_order = 'top_down',
+        follow_cursor = false,
       },
       docs = {
         auto_open = true,
--- a/lua/cmp/types/cmp.lua
+++ b/lua/cmp/types/cmp.lua
@@ -176,6 +176,7 @@ cmp.ItemField = {
 ---@class cmp.CustomEntriesViewConfig
 ---@field name 'custom'
 ---@field selection_order 'top_down'|'near_cursor'
+---@field follow_cursor boolean
 
 ---@class cmp.NativeEntriesViewConfig
 ---@field name 'native'
--- a/lua/cmp/view/custom_entries_view.lua
+++ b/lua/cmp/view/custom_entries_view.lua
@@ -153,9 +153,12 @@ custom_entries_view.open = function(self, offset, entries)
   height = height ~= 0 and height or #self.entries
   height = math.min(height, #self.entries)
 
+  local delta = 0
+  if not config.get().view.entries.follow_cursor then
+    local cursor_before_line = api.get_cursor_before_line()
+    delta = vim.fn.strdisplaywidth(cursor_before_line:sub(self.offset))
+  end
   local pos = api.get_screen_cursor()
-  local cursor_before_line = api.get_cursor_before_line()
-  local delta = vim.fn.strdisplaywidth(cursor_before_line:sub(self.offset))
   local row, col = pos[1], pos[2] - delta - 1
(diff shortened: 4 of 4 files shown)


Shougo:
CI seems failed.

mcauley-penney:
> CI seems failed.

If I'm reading the commit history correctly, that test has been failing since August 10th, so over two months. That gives me the impression that the error doesn't come from these two commits. If you believe it does and not from earlier commits, I'd like some help identifying where they come from.

Shougo:
> If I'm reading the commit history correctly, that test has been failing since August 10th, so over two months. That gives me the impression that the error doesn't come from these two commits. If you believe it does and not from earlier commits, I'd like some help identifying where they come from.

Oh....

So CI fix commit is needed.

mcauley-penney:
Just to be clear, you'd like a CI fix in this PR? I'd argue that that's outside of the scope of this particular PR, but I'm open to trying to fix it.

Shougo:
> Just to be clear, you'd like a CI fix in this PR? I'd argue that that's outside of the scope of this particular PR, but I'm open to trying to fix it.

I think it should be another PR.

mcauley-penney:
Okay, I can try. I am new to this code base, but I'd be open to trying to fix it and contributing to other things too. I am busy until Thursday morning. After that, I can take a look at that CI problem.

Shougo:
can you rebase the PR?
The CI seems fixed.

mcauley-penney:
@Shougo done. Thank you for orchestrating this

mcauley-penney:
@hrsh7th I've seen comments saying that you are feeling cautious about integrating new features because it will increase your maintenance burden, and you are busy. I would be happy to take ownership of this feature and its resulting issues if this PR were merged, similar to how Neovim itself has owners for new features.

Shougo:
The screenshot may help to understand the feature works.

mcauley-penney:
No problem at all, I'd be happy to provide some GIFs. To recap, the purpose of this PR is to allow the pmenu to follow the cursor as the user types. This feature is opt-in, and the current behavior remains the default.

## Current Behavior
Normally, the pmenu does not follow the user:

cmp-nocursor-customwin

## Additional Behavior
Using the custom window:
cmp-cursor-customwin

Using the native window:
cmp-cursor-nativewin


I have been using this since I made this PR, and I personally like it a lot. Some users may like this feature because it is like VSCode's autocomplete. Again, it is opt-in, so no one has to use it if they don't want to but it gives people another option.

hrsh7th:
IMO, it should be `custom_entries_view` only feature.

The native menu is a bit complicated, so I would like to exclude it from the list and annotate docs.

mcauley-penney:
@hrsh7th I have updated the PR so that the native menu is excluded and the docs do not mention it. Please let me know if there are any other changes you would like me to make.

hrsh7th:
LGTM. Thank you.

doc/cmp.txt
```diff
@@ -679,6 +679,13 @@ view.docs.auto_open~
 
   Specify whether to show the docs_view when selecting an item.
 
+                                                 *cmp-config.view.follow_cursor*
+view.follow_cursor*
```
bew, line 683: Hello, reading the rest of the final change it seems the help tags here are wrong, they should use `view.entries.follow_cursor` since the option is under `entries` 🤔

mcauley-penney: Good spot, thanks! I'll submit a fix today

mcauley-penney:
I've addressed my docs error as well as some other issues in the `view` docs in #1868

nyngwang:
> IMO, it should be `custom_entries_view` only feature.

What does this mean? Say I want to try the new option by this PR, then what should I do to ensure that it's only used with `custom_entries_view`?

nyngwang:
@hrsh7th:

> IMO, it should be `custom_entries_view` only feature.

What does this mean? Say I want to try the new option by this PR, then what should I do to ensure that it's only used with `custom_entries_view`?

mcauley-penney:
This option will not do anything when used with `name = 'native'. You don't necessarily have to avoid it, because it just won't do anything. 

That comment was left for me and was instructing me to make it so that the native view cannot use this option.

PR #1734, opened 2023-10-22
fix: remove usages from vim.lsp.util.parse_snippet

MariaSolOs:
This function is [now deprecated](https://github.com/neovim/neovim/pull/25733).

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -114,9 +114,6 @@ entry.get_word = function(self)
     local word
     if self:get_completion_item().textEdit and not misc.empty(self:get_completion_item().textEdit.newText) then
       word = str.trim(self:get_completion_item().textEdit.newText)
-      if self:get_completion_item().insertTextFormat == types.lsp.InsertTextFormat.Snippet then
-        word = vim.lsp.util.parse_snippet(word)
-      end
       local overwrite = self:get_overwrite()
       if 0 < overwrite[2] or self:get_completion_item().insertTextFormat == types.lsp.InsertTextFormat.Snippet then
         word = str.get_word(word, string.byte(self.context.cursor_after_line, 1), overwrite[1] or 0)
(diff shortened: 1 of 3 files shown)


MariaSolOs:
@hrsh7th I would really appreciate a look at this. The warning that these usages generate in nightly `neovim` are quite disruptive.

aarondill:
Not sure if this removes any capabilities, but as a nightly user, I would highly appreciate this PR being merged, as the deprecation messages actively interfere with my usage of nvim-cmp.

k1ng440:
@hrsh7th Please review this. Thank you

Shougo:
He seems busy...

mehalter:
Yeah it's also only been a couple days. It is open source software and they are maintaining this software for free and for the good of the community. It's probably better for people to not constantly ping them so frequently. If this PR is necessary for your set up, use that fork and branch in the meantime.

hrsh7th:
I'm not sure if "just deleting the existing process" is a reasonable solution, but if we don't merge it, there will likely be a big problem with the ecosystem, so let's merge it.

_google translated_

MariaSolOs:
@hrsh7th thank you for the review! That being said, I wish to provide some further context around the deprecation.

> I'm not sure if "just deleting the existing process" is a reasonable solution, but if we don't merge it, there will likely be a big problem with the ecosystem, so let's merge it.

[I was personally against](https://github.com/neovim/neovim/pull/25733#issuecomment-1773977761) the rapid changes around `parse_snippet`. However neovim is working really hard on building a better LSP API, and `parse_snippet` is an ill-designed function. Despite me being very annoyed at my completion configuration suddenly breaking like this, I agree with the rest of the core maintainers that this is the right thing to do.

The issue with this utility function is that it "parses" and formats using undefined heuristics. Its main purpose is to provide the `word` to use in vim completion items, but there are snippets with no good word replacement (e.g. postfix/multi-line snippets). This is a result of the lack of a 1:1 mapping between `vim` <-> LSP completion concepts, but from my experience this is a problem all editors that implement the LSP protocol encounter.

Neovim is doing its best to refactor these LSP APIs into robust editor interfaces that are fully compliant with the protocol while still retaining the magical features that `neovim` allows you to use. IMO we're doing a great job so far, and although this was a band-aid that really hurt ripping, I promise you that it will be worth it.

PR #1742, opened 2023-11-05
inline ghost text emulation for older neovim versions(v0.8.0-v0.10.0)

Sam-programs:
the emulation is working fine with treesitter and lsp but for the built-in syntax it might not be that accurate so i made it as an option and off by default until i figure out the issues with syntax

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -730,7 +730,7 @@ window.documentation.max_height~
 
                                             *cmp-config.experimental.ghost_text*
 experimental.ghost_text~
-  `boolean | { hl_group = string inline_emulation = boolean}`
+  `boolean | { hl_group: string, inline_emulation: boolean}`
   Whether to enable the ghost_text feature.
   inline_emulation is emulating inline ghost text for older neovim versions
 
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -730,8 +730,9 @@ window.documentation.max_height~
 
                                             *cmp-config.experimental.ghost_text*
 experimental.ghost_text~
-  `boolean | { hl_group = string }`
+  `boolean | { hl_group = string inline_emulation = boolean}`
   Whether to enable the ghost_text feature.
+  inline_emulation is emulating inline ghost text for older neovim versions
 
 ==============================================================================
 Config Helper                                                *cmp-config-helper*
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -732,7 +732,7 @@ window.documentation.max_height~
 experimental.ghost_text~
   `boolean | { hl_group: string, inline_emulation: boolean}`
   Whether to enable the ghost_text feature.
-  inline_emulation is emulating inline ghost text for older neovim versions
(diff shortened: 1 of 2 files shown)


Sam-programs:
i got it a lot better with memoization i still need to restructure the code a bit to handle comments and strings
https://github.com/Sam-programs/nvim-setup/blob/main/ghost_text_view.lua#L150

Sam-programs:
the memoization method would run into a lot of problems 
like operators being reused for multiple things in a language  
errors happening on the line where we cache somethings
not to mention strings and comments would be hard to handle because we would need to overwrite the memoized characters in the string/comment
```lua
local foo = bar -- a comment
```
the 2 --  are going to be loaded from the cache then we would detect we are a comment at the letter a then we would need to set the highlights to comments i don't think this is worth the effort   
if anyone has a better way to utilize memoization please help 

and i think it would be better to compute it only using the identifier splitting optimization
```lua
local foo = bar + 10
----- ---   ---   --
all of the underlined get 1 hilight as usual
= and + get inspected alone
```
this would result in many inspect calls but if we only compute it at first time and save the result for the line i think it won't matter

Sam-programs:
the issue with syntax is it crashes whenever i check if the highlight for it is cleared with `hl_iscleared`
so i picked the last highlight which seems to work best

Sam-programs:
never mind using syntax crashes noevim with error 139 regardless if i am checking it or not

doc/cmp.txt
```diff
@@ -730,8 +730,9 @@ window.documentation.max_height~
 
                                             *cmp-config.experimental.ghost_text*
 experimental.ghost_text~
-  `boolean | { hl_group = string }`
+  `boolean | { hl_group = string inline_emulation = boolean}`
```
wookayin, line 733:
```suggestion
  `boolean | { hl_group: string, inline_emulation: boolean}`
```

doc/cmp.txt
```diff
@@ -730,8 +730,9 @@ window.documentation.max_height~
...
   Whether to enable the ghost_text feature.
+  inline_emulation is emulating inline ghost text for older neovim versions
```
wookayin, line 735: Please make punctuation properly.

Which neovim version are you talking about --- that doesn't support inline text? This should be explicitly mentioned.

wookayin (changes requested):

doc/cmp.txt
Sam-programs: > Please make punctuation properly.
Sorry.

> Which neovim version are you talking about --- that doesn't support inline text? This should be explicitly mentioned.

v0.8.0+

Sam-programs:
Turns out the issue is inspect_pos is calling synstack inside on_line which causes neovim to crash,
probably because of this:
>  Note: doing anything other than setting extmarks is consideredg other than setting 
>  experimental.
I could use InsertCharPre to get the highlights for the line right before cmp starts,
but is there a better way to run code before cmp starts?

Sam-programs:
This was a bad idea, i only did it because i wasn't on nightly at the time of making this pr.

Issue #1698, opened 2023-09-09
Feature request: Support for Automatic Ghost Text and Manual Activation of Autocomplete Menu


mroavi:
I would love the feature that allows `ghost_text` to be automatically triggered with every keypress, and also provides the option to manually open the `autocomplete` menu.

Shougo:
Related https://github.com/hrsh7th/nvim-cmp/issues/1281 and https://github.com/hrsh7th/nvim-cmp/discussions/1361

Sam-programs:
i got a working version the only problem is it updates after u write a letter not with a keymap
it's also a bit hacky
```lua
local cmp = require('cmp')
local window = require('cmp.utils.window')
if old_cmp_win == nil then
   old_cmp_win = window.open
end
local cmp_menu_on = false
-- your field configs might add detection for that later
local cmp_fields_config = {
   -- this is the default
   'abbr', 'kind', 'menu' 
}
local cmp_window = nil
local function toggle_ghost_text()
   if cmp_menu_on then
      window.open = function(self, style)
         style.border = 'none'
         old_cmp_win(self, style)
         cmp_window = self.win
         vim.wo[cmp_window].winblend = 100
      end
      -- winblend with all spaces is opaque
      cmp.setup({
         formatting = {
            fields = {}
         }
      })
   else
      window.open = function(self,style)
         old_cmp_win(self,style)
         cmp_window = self.win
         vim.wo[cmp_window].winblend = 0
      end
      cmp.setup({
         formatting = {
            fields = cmp_fields_config
         }
      })
   end
   cmp_menu_on = not cmp_menu_on
end
vim.keymap.set('i', '<C-q>', toggle_ghost_text)
-- u need these 2 calls to overwrite the default functions
toggle_ghost_text() 
toggle_ghost_text() 
```

Sam-programs:
i got a better version that updates when the mapping is pressed
but it still doesn't hide the scrollbar which is a bit difficult
```lua
            local window = require('cmp.utils.window')
            if old_cmp_win == nil then
               old_cmp_win = window.open
            end
            local cmp_menu_on = false
            local cmp_window,cmp_border,cmp_w,cmp_h
            local buf = vim.api.nvim_create_buf(false,true)
            local emptybuf = vim.api.nvim_create_buf(false,true)
            local function toggle_ghost_text()
               if cmp_menu_on then
                  window.open = function(self, style)
                     cmp_border = style.border
                     cmp_w = style.width
[...]

Issue #1013, opened 2022-05-25
Hide the auto complete window completely


nparthas:
Hi, 

I was wondering if it was possible to completely hide the autocompletion window. I'd like to have something like this, which I saw was from [this post](https://stackoverflow.com/questions/53731133/show-the-autocomplete-match-as-preview-text-in-vim). 

I've tried changing some of the [completeopts](https://vimhelp.appspot.com/options.txt.html#%27completeopt%27) but taking a brief look through the code it seems like there are parts that are dependent of having `menu,menuone,noselect` as changing them in the plugin+nvim config led to some weird behaviour.

![image](https://user-images.githubusercontent.com/26097630/170199048-a87c8b5a-f280-4b40-8212-ce551d48a92b.png)

I saw the option for `ghost_text` which is what I'd like to use along with a hidden window, so that I can cycle through suggestions without having the popup take up space on the screen.

jonatan-branting:
I would love this feature. 

Given this was actually a thing, it would also be nice if it was possible to bring the completion menu up on demand, so that you aren't completely locked into just one line.

Shougo:
Hm.  If search mode or command line mode, statusline can be used for completion menu.
Please see.

https://github.com/hrsh7th/nvim-cmp/commit/124f1611f1b7cd9fb2892f0597881e007c99b652

Shougo:
https://github.com/hrsh7th/nvim-cmp/issues/406

`ghost_text` does not support hide the completion window.

oeyoews:
`autocomplete = true` is not support toggle, must comment this option in personal config file

hrsh7th:
The implementation should have been able to do this, but the current nvim-cmp is badly designed and difficult to implement. Sorry...

xiantang:
really good  feature

michaelfortunato:
Hope this is realized someday

PR #1750, opened 2023-11-12
Toggle completion menu

Sam-programs:
This pr gives the user the ability to hide all the cmp, by wrapping every `nvim_win*` function needed and storing the results before hiding the windows.
Fix:
https://github.com/hrsh7th/nvim-cmp/issues/1698
https://github.com/hrsh7th/nvim-cmp/issues/1013


dsully:
Would love to see this merged.

lua/cmp/view/ghost_text_view.lua
```diff
@@ -26,9 +26,13 @@ ghost_text_view.new = function()
...
-      return win == self.win
-    end,
-    on_line = function(_, _, _, on_row)
+      if self.extmark_id then
+        vim.api.nvim_buf_del_extmark(self.extmark_buf, ghost_text_view.ns, self.extmark_id)
+        self.extmark_id = nil
+      end
+      if win ~= self.win then
+        return
+      end
```
Sam-programs, line 35: This temporary because inline ghost text was broken in cmp when this was forked.

Sam-programs:
This still needs documentation, but i am not sure if the current config interface is good, i added 2 options.
```lua
window = {
      auto_hide = false,
      show_on_cursor_update = false,
}
```
`auto_hide` hides the windows after they opened until `require('cmp.utils.window').show_all()` is called or when the cursor moves in the menu if `show_on_cursor_update` is true.
However `show_on_cursor_update` doesn't work well because the cursor doesn't move with a pre-selected item,

sleepy-day:
Thanks for the work on this branch, I've been using it for a couple of months and forgot to check if it got merged in but it looks like it hasn't. Would it be possible to get this reviewed and merged in? Might need more work since it looks like it has merge conflicts and not sure if the creator of the branch is still around but if they're not I can clean up any issues and resubmit. Thank you.

Sam-programs:
> not sure if the creator of the branch is still around but if they're not I can clean up any issues

I am still here, but I don't plan to work on this pr anymore because I don't use cmp anymore. Sorry.

PR #1765, opened 2023-11-29
feat: sort source entries before slicing when using max_item_count

meeehdi-dev:
Hi,

When using the prop `max_item_count`, there is a known catch (sorting issue) from the doc:
```
sources[n].max_item_count~
  `number`
  The source-specific maximum item count option
  Note: This is applied before sorting, so items that aren't well-matched may be selected.
```

This PR makes it so the source entries are sorted before being sliced and fed to the full list of entries and then sorted again.
If the performance hit is too big, we could maybe add an optional boolean option such as `sort_before = true` along with `max_item_count`, to make sure the user still wants this sorting to happen even if it is slower.

EDIT:
Just had another thought... We could maybe avoid all this sorting X times per source + 1 time per group index, while still keeping the results accuracy, but it would bring some changes to how the final list is created. I'll try that later and maybe create another PR if it's worth it.

EDIT2:
Sorry for editing this PR once again but the first version felt wrong.
So this time what i did is just moving the whole `max_item_count` handling.
We fetch all the entries, we sort them, and at the end, depending on the source of the entry, we filter them out if they reached `max_item_count`.
I don't have that much experience with array manipulation in lua but i feel like the second commit would be way better performance-wise than the first one.

Thanks for your time.

### DIFF
--- a/lua/cmp/view.lua
+++ b/lua/cmp/view.lua
@@ -82,15 +82,26 @@ view.open = function(self, ctx, sources)
 
     -- create filtered entries.
     local offset = ctx.cursor.col
+    local group_entries = {}
+    local max_item_counts = {}
     for i, s in ipairs(source_group) do
       if s.offset <= ctx.cursor.col then
         if not has_triggered_by_symbol_source or s.is_triggered_by_symbol then
+          -- prepare max_item_counts map for filtering after sort.
+          if max_item_counts[s.name] == nil then
+            local max_item_count = s:get_source_config().max_item_count
+            if max_item_count ~= nil then
+              max_item_counts[s.name] = max_item_count
+            end
+          end
+
           -- source order priority bonus.
           local priority = s:get_source_config().priority or ((#source_group - (i - 1)) * config.get().sorting.priority_weight)
 
           for _, e in ipairs(s:get_entries(ctx)) do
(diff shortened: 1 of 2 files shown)


lua/cmp/view.lua
```diff
@@ -82,15 +82,24 @@ view.open = function(self, ctx, sources)
...
+            max_item_counts[s.name] = max_item_count
+          end
+
           -- source order priority bonus.
           local priority = s:get_source_config().priority or ((#source_group - (i - 1)) * config.get().sorting.priority_weight)
 
           for _, e in ipairs(s:get_entries(ctx)) do
             e.score = e.score + priority
-            table.insert(entries, e)
+            e.source_name = s.name -- store source name for filtering after sort.
```
hrsh7th, line 101: You don't need assign `e.source_name`.
You already can use `e.source.name` instead.

meeehdi-dev: I didn't notice the source was available in the entry. Thank you!

hrsh7th:
LGTM.

mrmeowski:
Cool feature - thank you! Should the documentation be updated to remove `Note: This is applied before sorting, so items that aren't well-matched may be selected.`?

PR #1776, opened 2023-12-14
feat(ci): update luarocks-tag-release to latest version

NTBBloodbath:
Hey!

Since the Luarocks releases CI was outdated, I have made some changes to update and improve it.

---

Ps: it would be nice if you made a new tag after this one is merge so that `nvim-cmp` is available in luarocks (the last tag was in 2022 so the CI hasn't really been used yet 😞)


hrsh7th:
thank you!

PR #1779, opened 2023-12-19
docs: fix invalid WindowConfig schema, should match to the docs

wookayin:
Problem: `cmp.WindowConfig` was defined twice, and not consistent with docs |cmp.txt|.

Solution: Introduce `cmp.CompletionWindowOptions` and `cmp.DocumentationWindowOptions`.
Make fields of these two class consistent with |cmp-config.window.*|


hrsh7th:
perfect. thank you.

PR #1805, opened 2024-01-23
feat: support completion in terminal mode

### LINKED ISSUE #1219: Question: Completion in nvim's terminal window
Hi!

First: Thank you very much for the different completion/ source plugins @hrsh7th! The stuff is really helpful!

My question:
Is there any chance/ trick how I could achieve that `nvim-cmp` triggers in the internal terminal window?

Use cases:

* Shell script snippets in zsh shell
* Python snippets in IPython shell or Python shell
* In general stuff provided by the [`cmp-buffer`](https://github.com/hrsh7th/cmp-buffer)

That would be really awesome!

many thanks and all best,
Jamil

willothy:
An attempt / POC of making completion work in terminal mode. Currently working with cmp-buffer.

closes #1219 (maybe)

https://github.com/hrsh7th/nvim-cmp/assets/38540736/7403a798-9bc9-4380-97f3-373a269a87ae

### DIFF
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -387,7 +387,7 @@ core.confirm = function(self, e, option, callback)
       table.insert(keys, keymap.backspace(ctx.cursor_before_line:sub(e:get_offset())))
       table.insert(keys, string.sub(e.context.cursor_before_line, e:get_offset()))
       feedkeys.call(table.concat(keys, ''), 'in')
-    else
+    elseif not api.is_terminal_mode() then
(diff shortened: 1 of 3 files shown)


willothy:
Closing as I didn't end up finding this particularly useful in terminal mode (builtin zsh completion works better for now), but was still an interesting attempt. Would maybe be more useful if it would work with other sources than buffer and path since builtin completions do those very well already. Copilot could be interesting, but I wasn't able to get that working.

PR #1813, opened 2024-02-03
feat: avoid global variable use

### LINKED ISSUE #1810: Please consider not using global variables
I've recently noticed many of the plugins I use make unnecessary variables in the global namespace. I see that `cmp` is using it for a [function call](https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/utils/feedkeys.lua#L27). I wanted to ask, if it would be possible to replace it with something like `<Cmd>lua require'cmp.utils.feedkeys'.run(%s)<CR>` moving the [global function](https://github.com/hrsh7th/nvim-cmp/blob/main/lua/cmp/utils/feedkeys.lua#L45) to the module, where it comes from.

JosefLitos:
fixes #1810

### DIFF
--- a/lua/cmp/utils/feedkeys.lua
+++ b/lua/cmp/utils/feedkeys.lua
@@ -24,7 +24,7 @@ feedkeys.call = setmetatable({
     if callback then
       local id = misc.id('cmp.utils.feedkeys.call')
       self.callbacks[id] = callback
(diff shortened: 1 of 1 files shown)


hrsh7th:
LGTM. Thank you.

PR #1824, opened 2024-02-12
fix: small windows no longer constantly throw errors

### LINKED ISSUE #827: Using nvim-cmp with firenvim cause problem
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp#faq) and it didn't resolve my problem.

### Issues

### Neovim Version

0.7.0dev

### Minimal reproducible config

https://github.com/tin-pham/NeoVimLua

It inside lua/options/cmp.lua

### Description

It show this error when I use firenvim on text box on the browser (markdown textbox like reddit, github work fine. Other textbox not markdown will cause this)
image

### Steps to reproduce

1. Use nvim-cmp with firenvim
2. If the textbox is not markdown (facebook, googletranslate) it will cause this

### Expected behavior

1. It just like the image above

### Actual behavior

It will just writing normal like this without that text being annoy pop up all over again
image

### Additional context

I have open issue on the github site of the firenvim author, and have some discussion with the admin

I can turn off nvim-cmp and start using firenvim on the web but when I start coding, it can be annoy to turn it on again, so maybe my config have something that cause this problem??

https://github.com/glacambre/firenvim/issues/1330

### LINKED ISSUE #1759: `custom_entries_view` throws an error when the window is too small
### FAQ

### Announcement

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
  },

    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}
EOF

lua << EOF

  capabilities = capabilities,
}
EOF

```

### Description

`...d/vim-plug/nvim-cmp/lua/cmp/view/custom_entries_view.lua:210: Expected Lua number`

This error keeps occurring every single time when cmp tries to open the completion menu to complete the word.

### Steps to reproduce

1. Use the configuration.
2. Make the terminal window tiny, as shown below:
image
3. The error will keep occurring when there is more than 1 completion item.

### Expected behavior

No error occurs and I get a tiny menu that has 1 or 2 words for the auto-completion. It used to work like that until something changed.

### Actual behavior

`...d/vim-plug/nvim-cmp/lua/cmp/view/custom_entries_view.lua:210: Expected Lua number`

This error keeps occurring and it's nearly impossible to type anything.

### Additional context

While the terminal window being that tiny may seem unreasonable, this actually happens a lot when using Firenvim on small text areas such as search bars and text boxes for chat apps.

hankertrix:
Fixes #827 and #1759

The `entries_win:open` function will fail when either the width or the height of the window is less than 1, which will result in the `nvim_win_set_cursor` being passed a `nil` value for `self.entries_win.win` instead of a number, which causes an error to constantly appear when the window is small. Hence, a guard clause to check for whether `self.entries_win.win` is `nil` is added to stop the error from occurring.

Also fixed the rarer 'height' must be a positive Integer error caused by the `window.update` function, within the clause to draw the scrollbar background. This error happens on small windows when pressing and holding down the backspace key to delete a lot of characters at once. The `nvim_open_win` function call to create the scrollbar background throws the error when `self.style.height` is `0`. Hence, an additional check is added alongside the `info.scrollable` check which also skips drawing the scrollbar thumb as it is not needed when the height is `0` and will result in a weird scrollbar thumb floating some distance away from the text when holding down backspace to delete a lot of characters.

### DIFF
--- a/lua/cmp/utils/window.lua
+++ b/lua/cmp/utils/window.lua
@@ -133,7 +133,7 @@ end
 ---Update
 window.update = function(self)
   local info = self:info()
-  if info.scrollable then
+  if info.scrollable and self.style.height > 0 then
     -- Draw the background of the scrollbar
 
     if not info.border_info.visible then
--- a/lua/cmp/view/custom_entries_view.lua
+++ b/lua/cmp/view/custom_entries_view.lua
@@ -212,7 +212,14 @@ custom_entries_view.open = function(self, offset, entries)
     border = completion.border,
     zindex = completion.zindex or 1001,
   })
-  -- always set cursor when starting. It will be adjusted on the call to _select
+
+  -- Don't set the cursor if the entries_win:open function fails
+  -- due to the window's width or height being less than 1
+  if self.entries_win.win == nil then
+    return
+  end
+
+  -- Always set cursor when starting. It will be adjusted on the call to _select
   vim.api.nvim_win_set_cursor(self.entries_win.win, { 1, 0 })
   if preselect_index > 0 and config.get().preselect == types.cmp.PreselectMode.Item then
     self:_select(preselect_index, { behavior = types.cmp.SelectBehavior.Select, active = false })


hrsh7th:
Thank you!

Issue #1818, opened 2024-02-08
Main suggestion gets stuck in terminal, after changing focus to another window


blauweb:
Hi,

If I change the window once a suggestion has appeared next to my cursor, a ghost of that suggestion stays in that line of code and doesn't go away until I reset vim.

The issue happens specifically when I switch the focus to a different window (inside nvim) by mouse, while in insert mode; this does not happen if I change focus to a different app, or if I exit insert mode first and then change the focus.

This is not related to: #201 , the data stuck is something else.

Environment:
Ubuntu 22.04
Gnome 42
Neovim with Lazyvim.

![image](https://github.com/zbirenbaum/copilot.lua/assets/38893376/af38a35e-36e7-4330-bf1e-9778739ee4b6)

PR #1826, opened 2024-02-17
fix(ghost_text): make sure to the delete the mark for the same buffer

Sam-programs:
Fix: https://github.com/hrsh7th/nvim-cmp/issues/1818


hrsh7th:
Thank you!

Issue #1832, opened 2024-02-21
Crash plugin while navigate in menu snippets LuaSnip (nvim-cmp/lua/cmp/utils/str.lua:81: bad argument #1 to byte (string expected, got table) 


Forsigg:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```lua
local cmp = require 'cmp'
local luasnip = require 'luasnip'


local has_words_before = function()
  unpack = unpack or table.unpack
  local line, col = unpack(vim.api.nvim_win_get_cursor(0))
  return col ~= 0 and vim.api.nvim_buf_get_lines(0, line - 1, line, true)[1]:sub(col, col):match("%s") == nil
end

local cmp_kinds = {
  Text = '  ',
  Method = '  ',
  Function = '  ',
  Constructor = '  ',
  Field = '  ',
  Variable = '  ',
  Class = '  ',
  Interface = '  ',
  Module = '  ',
  Property = '  ',
  Unit = '  ',
  Value = '  ',
  Enum = '  ',
  Keyword = '  ',
  Snippet = '  ',
  Color = '  ',
  File = '  ',
  Reference = '  ',
  Folder = '  ',
  EnumMember = '  ',
  Constant = '  ',
  Struct = '  ',
  Event = '  ',
  Operator = '  ',
  TypeParameter = '  ',
}

cmp.setup({
    snippet = {
        -- REQUIRED - you must specify a snippet engine
        expand = function(args)
            -- vim.fn["vsnip#anonymous"](args.body) -- For `vsnip` users.
            luasnip.lsp_expand(args.body) -- For `luasnip` users.
            -- require('snippy').expand_snippet(args.body) -- For `snippy` users.
            -- vim.fn["UltiSnips#Anon"](args.body) -- For `ultisnips` users.
        end
    }, 
    formatting = {
        format = function(_, vim_item)
            vim_item.kind = (cmp_kinds[vim_item.kind] or '') .. vim_item.kind
            return vim_item
        end,
    },
    window = {
        completion = cmp.config.window.bordered(),
        documentation = cmp.config.window.bordered()
    },
    mapping = {
        
        ["<Tab>"] = cmp.mapping(function(fallback)
[...]

PR #1833, opened 2024-02-21
fix: issue#1832

Forsigg:
Add type check for `item.detail` that contains description, In that case while navigating in menu for select snippet `item.detail` contains table, but expected string

#1832

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -434,6 +434,10 @@ entry.get_documentation = function(self)
   local item = self:get_completion_item()
 
   local documents = {}
+  
+  if type(item.detail) == 'table' then
+    item.detail = item.detail[1]
+  end
 
   -- detail
   if item.detail and item.detail ~= '' then


fivepixels:
@hrsh7th, can you please merge this as quickly as possible? I got the same error as him(#1832). When I added the codes that he wrote into my local nvim cmp codes, it worked without making any errors. Please check it and merge it. @Forsigg THank you!

hrsh7th:
IMO, the item.detail as table is not compliant fir LSP spec. So I tgink it should be fixed in source side.

rbjorklin:
I also just hit this and the commit in this PR works as a work-around.

@hrsh7th What is the "source side" in this scenario? Should this bug be reported against another project? If so, which one?

Shougo:
> @hrsh7th What is the "source side" in this scenario? Should this bug be reported against another project? If so, which one?

If it is source side problem, the source should check the `item.detail` instead.

PR #1860, opened 2024-03-24
Use `get_word` so that the word is the same as in `core.confirm`

luozhiya:
## Problem

Build a `CompletionResponse` in the following way:
```lua
local label = cursor_before_line .. suggestions
-- CompletionResponse
{
  label = label,
  word = label,
  textEdit = {
    range = {
      start = { line = line, character = character },
      ['end'] = { line = line, character = character },
    },
    newText = suggestions,
  },
}
```
When `cursor_before_line` is `e` and `suggestions` is `mail`, ghost_text will show `ail` and will be missing a character `m`.
However, when you preview it from `<Tab>`, the `mail` is displayed correctly.

## Solution

Use `get_word` so that the word is the same as in `core.confirm`

### DIFF
--- a/lua/cmp/view/ghost_text_view.lua
+++ b/lua/cmp/view/ghost_text_view.lua
@@ -74,7 +74,7 @@ end
 ---  This function calculates the bytes of the entry to display calculating the number
 ---  of character differences instead of just byte difference.
 ghost_text_view.text_gen = function(self, line, cursor_col)
-  local word = self.entry:get_insert_text()
+  local word = self.entry:get_word()
   word = str.oneline(word)
   local word_clen = vim.str_utfindex(word)
   local cword = string.sub(line, self.entry:get_offset(), cursor_col)


liskin:
This only works if there's a `word` field in the CompletionResponse, but it often isn't—certainly not if the completion comes from cmp-omni.

I've came up with another attempt to fix the same issue:
```diff
diff --git a/lua/cmp/view/ghost_text_view.lua b/lua/cmp/view/ghost_text_view.lua
index 19c9513..1af7ad9 100644
--- a/lua/cmp/view/ghost_text_view.lua
+++ b/lua/cmp/view/ghost_text_view.lua
@@ -88,6 +88,12 @@ ghost_text_view.text_gen = function(self, line, cursor_col)
   if self.entry:get_completion_item().insertTextFormat == types.lsp.InsertTextFormat.Snippet then
     word = tostring(snippet.parse(word))
   end
+
+  local irange = self.entry:get_insert_range()
+  if irange and self.entry.source_insert_range.start.line == irange.start.line then
+    word = string.sub(line, self.entry:get_offset(), irange.start.character) .. word
+  end
+
   local word_clen = vim.str_utfindex(word)
   local cword = string.sub(line, self.entry:get_offset(), cursor_col)
   local cword_clen = vim.str_utfindex(cword)

```

Since `cword` is set as `string.sub(line, self.entry:get_offset(), cursor_col)`, `word` also needs to start with the characters between `self.entry:get_offset()` and the start of the insert range, if the insert range doesn't start at offset. Which it doesn't for completions coming from `python3complete#Complete` via `cmp-omni`, those look like this:

```lua
  {                                                                                                                                                              
    label = "service_command(service_name)",
    labelDetails = {},
    textEdit = {
      newText = "ice_command(",
      range = {
        ["end"] = {
          character = 12,
          line = 456
        },
        start = {
          character = 12,
          line = 456
        }
      }
    }
  }
```

This is after typing "serv" and triggering completions. `newText` doesn't contain the "serv", it just appends stuff to the end.

(Also, with such textEdits, replacing is completely broken because of the silly heuristic introduced in https://github.com/hrsh7th/nvim-cmp/pull/1177)

PR #1861, opened 2024-03-24
feat: make winblend configurable

ofseed:
Set cmp window's `winblend` to `pumblend` makes it consistent with neovim itself, but there is a case:

image

As you can see above, some of the nerd icons on the left side of the window are small, that's because terminals render nerd icons as two characters wide if there is a space on their right. If `pumblend` is set, the right of the nerd icon will be a character under the window instead of a space. So I think a positive value of `winblend` is not suit for cmp in some case.

In current implementation, there's no way to make cmp's winblend different from other popup windows, if I want to change cmp, I must change them all. So I recommend making this option configurable.

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -698,6 +698,12 @@ window.{completion,documentation}.winhighlight~
   Specify the window's winhighlight option.
   See |nvim_open_win|.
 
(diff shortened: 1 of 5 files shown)


hrsh7th:
make sense. Thank you.

PR #1868, opened 2024-03-31
docs: update and fix `view` documentation

mcauley-penney:
- correct view.follow_cursor to view.entries.follow_cursor
- mention that view.entries.follow_cursor is custom view only
- add missing view.entries.selection_order option
- mention the docs class in list of classes nested under view class
- trim errant whitespace

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -257,7 +257,7 @@ NOTE: `<Cmd>lua require('cmp').complete()<CR>` can be used to call these functio
   `behavior=cmp.ConfirmBehavior.Insert`: inserts the selected item and
     moves adjacent text to the right (default).
   `behavior=cmp.ConfirmBehavior.Replace`: replaces adjacent text with
-    the selected item.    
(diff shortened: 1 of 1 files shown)


doc/cmp.txt
```diff
@@ -686,12 +687,21 @@ view.docs.auto_open~
 
   Specify whether to show the docs_view when selecting an item.
 
-                                                 *cmp-config.view.follow_cursor*
-view.follow_cursor*
+                                                *cmp-config.view.entries.follow_cursor*
```
hrsh7th, line 690: it should be selection_order?

mcauley-penney: I've added selection_order above this as a separate entry. `config.view.entries.follow_cursor` is the proper name for the option added by my PR yesterday, but `config.view.entries.selection_order` is another option that is already available to the user but is undocumented in this section as of right now.

hrsh7th:
Thank you!

PR #1896, opened 2024-04-21
perf: don't eagerly load vim.snippet

tomtomjhj:
Evaluating `vim.snippet` loads many modules, which takes about 1ms.


mehalter (approved):
This is much better than what I did in the original PR to add this. Thanks so much!

hrsh7th:
thank you

PR #1911, opened 2024-05-06
Do not use sync code in InserEnter

### LINKED ISSUE #1908: cmp.core:prepare() is expansive and calls for every InsertEnter
https://github.com/hrsh7th/nvim-cmp/blob/8f3c541407e691af6163e2447f3af1bd6e17f9a3/lua/cmp/init.lua#L333
It calls `cmp.core:prepare()` for every InsertEnter and it takes 20ms to run that prepare function, which makes `o` become laggy, it is bad.

xzbdmw:
close #1908 
debounce_next_tick_by_keymap is a sync call, and this is the only place it is called.
This change makes o faster.

### DIFF
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -330,7 +330,7 @@ local on_insert_enter = function()
   end
 end
 autocmd.subscribe({ 'CmdlineEnter' }, async.debounce_next_tick(on_insert_enter))
-autocmd.subscribe({ 'InsertEnter' }, async.debounce_next_tick_by_keymap(on_insert_enter))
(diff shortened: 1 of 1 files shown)


mortezadadgar:
@hrsh7th anything stop this pr from being merged? like OP I recorded a profile that is showing `cmp.visible` is causing long delays 
[profile.json](https://github.com/user-attachments/files/19245850/profile.json)

https://github.com/user-attachments/assets/167d28fb-51c2-4c22-8674-e7ae9c8fd8fa

mortezadadgar:
yeah sorry it's unrelated

PR #1914, opened 2024-05-08
Docs: fix README.md by adding setup for cmp-git

ilan-schemoul:


ilan-schemoul:
We need setup() but I commented everything as cmp-git is not added in the dependencies in the example so calling setup would fail by default

hrsh7th:
Thank you.

Issue #1919, opened 2024-05-12
Unexpected behavior with cmp.abort() and cmp.close() when autocomplete = false.


Ph7s1c:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```lua
local cmp = require('cmp')
local cmp_select = { behavior = cmp.SelectBehavior.Select }

cmp.setup({
    mapping = {
        ['<Tab>'] = cmp.mapping(function(fallback)
            if cmp.visible() or has_words_before() then
                cmp.complete()
        	    cmp.confirm({ select = true })
            else
                fallback()
            end
        end, {"i","s","c",}),
        ['<Up>'] = cmp.mapping.select_prev_item(cmp_select),
        ['<Down>'] = cmp.mapping.select_next_item(cmp_select),
        ['<C-Space>'] = cmp.mapping(function()
            if cmp.visible() then
                cmp.abort()
            else
                cmp.complete()
            end
        end, {"i","s","c",}),
    },
    completion = {
        autocomplete = false,
    },
    sources = cmp.config.sources({
        { name = 'treesitter' },
        { name = 'calc' },
        { name = 'buffer'},
    }, {
        { name = 'path',
          option = {
            trailing_slash = true},
        },
    })
})
```

### Description

https://github.com/hrsh7th/nvim-cmp/assets/59641969/6ec6e92d-7d0a-4eda-bfd0-2f8cbf1b979e

When calling cmp.abort() or cmp.close(), with manual completion enabled, the completion menu re-appears after typing more characters. 

### Steps to reproduce

Replication steps:

Spawn a completion menu with ctrl-space.
Abort the menu with ctrl-space again
(or a different bind which doesn't make any difference).

Type more characters or delete some characters.

The menu pops up again, which shouldn't happen, since completion has been
aborted, and autocomplete=false.

### Expected behavior

After completion window is aborted, the window should only re-appear once I call cmp.complete() again, rather than when I type more characters.

### Actual behavior

Completion window is initially closed, but typing more characters re-opens the window.

### Additional context

[...]

PR #1920, opened 2024-05-12
fix: unexpectedly re-open a completion window when setting `autocomplete = false`

### LINKED ISSUE #1251: With autocomplete=false, cmp menu will still auto appear after aborting if text under cursor matches a suggestion
### FAQ

### Announcement

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
local has_words_before = function()
  local line, col = unpack(vim.api.nvim_win_get_cursor(0))
  return col ~= 0 and vim.api.nvim_buf_get_lines(0, line - 1, line, true)[1]:sub(col, col):match("%s") == nil
end
cmp.setup {
  completion = {
    autocomplete = false,
  },

  mapping = {
    -- https://github.com/hrsh7th/nvim-cmp/wiki/Example-mappings#luasnip
    ["<Tab>"] = cmp.mapping(function(fallback)
      if cmp.visible() then
        cmp.select_next_item()
        cmp.complete({
          reason = cmp.ContextReason.Auto,
        })
      else
        fallback()
      end
    end, { "i", "s" }),

    ["<C-e>"] = cmp.mapping.abort(),
  },

    { name = "buffer" },
  }),
}
EOF
```

### Description

With autocomplete set to false, if the cmp menu is manually triggered and then manually closed (eg, via `cmp.mapping.abort()` or `cmp.mapping.close()`), the cmp menu will re-appear while the cursor is within the original completion area.

 

https://user-images.githubusercontent.com/29129/197155715-6953cf9f-9e05-4fa3-b199-c5f55d1fe374.mov

### Steps to reproduce

1. Using the provided config
2. Start typing a word that will have buffer completions
3. Hit `<Tab>` to trigger cmp menu
4. Hit `<C-e>` to abort the menu
5. Continue typing, matching at least one of the suggestions from the cmp menu
6. Cmp menu reappears automatically

### Expected behavior

Completion menu only appears when manually triggered if autocomplete is set to false.

### Actual behavior

Completion menu reappears after being aborted as long as text under cursor continues to match suggestions.

### Additional context

_No response_

kjuq:
Fix #1251 and #1919. This comment was very helpful for me. (https://github.com/hrsh7th/nvim-cmp/issues/1251#issuecomment-1365229958)

### DIFF
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -120,6 +120,7 @@ cmp.abort = cmp.sync(function()
   if cmp.core.view:visible() then
     local release = cmp.core:suspend()
     cmp.core.view:abort()
+    cmp.core:reset()
     vim.schedule(release)
     return true
   else


hrsh7th:
thank you!

PR #1930, opened 2024-05-16
Move to non deprecated function calls

augustocdias:
`vim.tbl_islist` will be deprecated on 0.12. This make use of their replacement. I'm not sure what's the minimum version this function is available though...

### DIFF
--- a/lua/cmp/utils/misc.lua
+++ b/lua/cmp/utils/misc.lua
@@ -1,5 +1,7 @@
 local misc = {}
 
+local islist = vim.islist or vim.tbl_islist
+
 ---Create once callback
 ---@param callback function
 ---@return function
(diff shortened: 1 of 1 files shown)


seblj:
You can just do

```lua
local islist = vim.islist or vim.tbl_islist
```

And then use that, so that it won't stop working for the ones that haven't updated yet

augustocdias:
> You can just do
> 
> ```lua
> local islist = vim.islist or vim.tbl_islist
> ```
> 
> And then use that, so that it won't stop working for the ones that haven't updated yet.
> 
> There is also a warning for `vim.tbl_add_reverse_lookup` if you want to tackle that as well

Good point. Just did that. And about the other warning, I'm not sure what it should replace it with... There's nothing in the deprecation warning...

seblj:
> And about the other warning, I'm not sure what it should replace it with... There's nothing in the deprecation warning...

You can probably find some examples in this where it was deprecated:

https://github.com/neovim/neovim/commit/e52c25b7617ac6401b080f76b0e227161dfef230

seanbreckenridge:
> There is also a warning for vim.tbl_add_reverse_lookup if you want to tackle that as well

Seems https://github.com/hrsh7th/nvim-cmp/pull/1931 fixes that

AThePeanut4:
@augustocdias If you wouldn't mind adding the change here, this would probably be the easiest place to fix https://github.com/hrsh7th/nvim-cmp/pull/1931#issuecomment-2117054998

hrsh7th:
thank you

PR #986, opened 2022-05-13
fix: use scratch buffer and lhs for keymap.normalize()

zeertzjq:
Fix https://github.com/neovim/neovim/issues/18303

This can be slower, I don't know how frequently this function is called as I don't use this plugin.

PR #1935, opened 2024-05-17
perf: use builtin for key normalization

tomtomjhj:
After https://github.com/hrsh7th/nvim-cmp/pull/986, @zeertzjq added the `keytrans()` function for translating byte rep of keys to key notation form in https://github.com/neovim/neovim/commit/907fc8ac373226556b84c2fdc4fe26525bbdb2c4. This can be used for normalization. 

This reduces running time of `prepare()` from about 16ms to 6ms on my setup. 

btw I' curious about why `keymap.t` manually replaces `<` instead of using `nvim_replace_termcodes` (https://github.com/hrsh7th/nvim-cmp/pull/720/commits/f2f6dce0cc708ebea904ed471849e1d62a19cf96). `nvim_replace_termcodes` is slightly faster.

### DIFF
--- a/lua/cmp/utils/keymap.lua
+++ b/lua/cmp/utils/keymap.lua
@@ -16,7 +16,9 @@ end
 ---Normalize key sequence.
 ---@param keys string
 ---@return string
-keymap.normalize = function(keys)
+keymap.normalize = vim.fn.has('nvim-0.8') == 1 and function(keys)
+  return vim.fn.keytrans(keymap.t(keys))
+end or function(keys)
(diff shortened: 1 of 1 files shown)


hrsh7th:
thank you.

tomtomjhj:
Thta seems to be nvim bug. reported https://github.com/neovim/neovim/issues/30887

dav-cho:
After this commit, I'm seeing a slight delay when typing `<` in insert mode. #2069 does not fix this. Any ideas on how to address?

PR #1955, opened 2024-06-08
Multiline snippet ghost text with dynamic window position flip

xzbdmw:


michaelbrusegard:
This is so awesome! I will test it shortly! Can you share how you have set up CMP @xzbdmw . I want to try and avoid making a mistake in the setup on my part. I was using Copilot.lua with the cmp extension for my ghost text

xzbdmw:
Hey I think you don’t need to change anything. should work out of box

cybergaz:
please add an option to toggle between multiline and singleline ghost text, not everyone like to have multiline previews.

michaelbrusegard:
@xzbdmw It works really well and swaps every time there is a multiline ghost text! My completion menu does not seem to follow the mouse as much as yours and I am not completely sure about the hiding of the menu when there is a space (but keeping the ghost text). If you want to take a look at my cmp config: https://github.com/michaelbrusegard/dotfiles/blob/25f81267d8f3d9253153b54bf92dd6e757ac089a/macos/.config/nvim/init.lua#L551

Maybe you have some feedback or suggestions, I am very new to neovim btw
Thanks for the great work

xzbdmw:
@cybergaz That shoud be easy
@michaelbrusegard could you use a gif to show that

lua/cmp/view.lua
```diff
@@ -295,8 +295,7 @@ view.on_entry_change = async.throttle(function(self)
   else
     self.docs_view:close()
   end
-
-  e = e or self:get_first_entry()
+  e = e or self:get_selected_entry()
```
xzbdmw, line 298: I think this makes sense, if user specify 'noselect', don't show ghost text

michaelbrusegard:
https://github.com/hrsh7th/nvim-cmp/assets/56915010/5f5dcce7-968c-47a3-bf0a-d5a462624acd

This is how it looks, but the menu doesnt hide on space (while keeping the ghost text) and it doesnt move with the cursor. It may be my config is wrong for all I know

xzbdmw:
> but the menu doesnt hide on space

That’s because copilot suggestion contains space, so it just matches

>  it doesnt move with the cursor

you can try 
require("cmp").setup({
view = {
entries = {
follow_cursor = true
}
}
})

michaelbrusegard:
@xzbdmw Thank you, that worked with the cursor. Are there any convenient way to make the keybinds. I feel like when I have the completion complete with return or tab it can be annoying if I actually want to use any of those

michaelbrusegard:
@xzbdmw `follow_cursor=true` only works when the completion menu shows up below the line. When it shows up above it is placed at the start.
Screenshot 2024-08-27 at 17 44 10

xzbdmw:
Have you updated to the latest commit? works as expect to me.

https://github.com/user-attachments/assets/07940428-8f38-4a6e-920b-6def203d85dc

michaelbrusegard:
You are right... I had removed the follow cursor option from my config. Probably by accident once

xzbdmw:
@mikecsmith If the completion menu can't be fit in area that above the cursor, the menu still stays below, do you have some way to reproduce that issue?

https://github.com/user-attachments/assets/9b1fa0da-0555-4427-a7fd-ae908bf9979e

barnacker:
Currently running off this branch with high satisfaction, combined with copilot

xzbdmw:
I've rebased with master that has perf improvement.

vex9z7:
waiting for merge, we need this feature

xzbdmw:
I've integrate with [mini-snippets](https://github.com/echasnovski/mini.snippets) with [a source](https://github.com/xzbdmw/cmp_mini_snippets), so that snippets defined by mini.snippets has the same look as lsp snippets in this pr.

There is a video in https://github.com/echasnovski/mini.nvim/issues/1428#issuecomment-2566571295

wsdjeg:
@xzbdmw Hi, thanks for this PR, can you fix the conflicts in this pr?

Issue #1887, opened 2024-04-16
feature-request: allow more control over completion highlight


xzbdmw:
Zed:
<img width="616" alt="截屏2024-04-17 02 45 02" src="https://github.com/hrsh7th/nvim-cmp/assets/97848247/36abf371-46cb-4b03-98b0-5ddb3559217f">

They achieve these beautiful colors by regex substitute [`completion.details`](https://github.com/zed-industries/zed/blob/52591905fb8b399401082f7854be4a0fa5677627/crates/languages/src/rust.rs#L181) from language server and apply highlight to them.Currently cmp apply different highlight at a fix width, would be great to expose variable-size highlight(the same format as nvim_buf_set_extmark?) interface to users(and full lsp response message).

Edit: they are just concating detail items to fake it into valid treesitter grammer and apply highlight directly using [treesitter query](https://github.com/zed-industries/zed/blob/e25f0dfb0a2f76829a2fdd32b600da57ce3c747d/crates/language/src/language.rs#L1272). The fake process is fairly easy and already can be done by formatting config. I assume cmp only needs an option to apply treesitter highlights or not for completion items.

xzbdmw:
quick demo
<img width="923" alt="截屏2024-04-18 09 19 42" src="https://github.com/hrsh7th/nvim-cmp/assets/97848247/533eca23-8c76-4bba-84d5-085a01867c84">
![image](https://github.com/hrsh7th/nvim-cmp/assets/97848247/80de129b-4765-40e6-a7d7-c2f5658f58b2)

Moerliy:
Would love to see this happen.

PR #1972, opened 2024-06-30
feat: allow more completion highlight control

dmun:
Related: #1887

This PR allows more highlight control by allowing `vim.CompletedItem.{field}_hl_group` values to be of type `table` with the following structure:

```lua
-- inside the cmp.FormattingConfig.format method
vim_item.abbr_hl_group = {
    { '@function', range = { 0, 5 } },
    { 'Comment', range = { 5, 10 } },
}
```

### Simple
A simple use case of this is having different highlights for function parameters:

image

<details>
<summary>Code</summary>

```lua
local format = function(entry, vim_item)
    local kind = entry:get_kind()

    if vim.tbl_contains({ 2, 3 }, kind) then -- if kind is a Function (2) or Method (3)
        local start_index, end_index = vim_item.abbr:find('%(.*')
        if start_index then
            vim_item.abbr_hl_group = {
                { '@function', range = { 0, start_index - 1 } },
                { 'Comment', range = { start_index - 1, end_index } },
            }
        else
            vim_item.abbr_hl_group = '@function'
        end
    end

    return vim_item
end
```

</details>

### Treesitter-based
You can also have more complex highlights by using treesitter for example:

image

<details>
<summary>Code (borrowed from @xzbdmw)</summary>

```lua
format = function(entry, vim_item)
    local highlights = {}

    -- you will likely not want to get this query every single time for performance but this is an example
    local query = vim.treesitter.query.get(vim.bo.filetype, 'highlights')

    local success, parser = pcall(vim.treesitter.get_string_parser, str, vim.bo.filetype)
    if success then
        local tree = parser:parse(true)[1]
        local root = tree:root()
        for id, node in query:iter_captures(root, str, 0, -1) do
            local name = '@' .. query.captures[id] .. '.' .. vim.bo.filetype
            local hl = vim.api.nvim_get_hl_id_by_name(name)
            local range = { node:range() }
            local _, nscol, _, necol = range[1], range[2], range[3], range[4]

            table.insert(highlights, { hl, range = { nscol, necol } })
        end
    end

    vim_item.abbr_hl_group = highlights

    return vim_item
end,
```
</details>

## Notes
LSP doesn't specify exactly how to use the label fields of a `CompletionItem`, as a result the usage of these values differ a lot between LSP implementations (see table below). This makes getting syntactically correct function signatures inconsistent across different LSP implementations. 

| LS | label | detail | labelDetails.detail | labelDetails.description |
| ------------- | -------------- | -------------- | -- | -- |
| lua_ls | `require(modname)` | `function` | - | - |
| zls | `addManyAt` | `fn (self: *Self, index: usize, count: usize) Allocator.Error![]T` | `(index: usize, count: usize)` | `Allocator.Error![]T` |
| rust-analyzer | `find_map(…)` | `fn(&mut self, F) -> Option<B>` | ` (as Iterator)` | `fn(&mut self, F) -> Option<B>` |

I don't think nvim-cmp should be responsible for properly formatting/highlighting the function signatures. Instead, that logic should be configurable by the user, or delegated to some other plugin in similar style of [lspkind-nvim](https://github.com/onsails/lspkind.nvim).

Suggestions are very welcome!

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -201,7 +201,7 @@ end
 ---Return view information.
 ---@param suggest_offset integer
 ---@param entries_buf integer The buffer this entry will be rendered into.
----@return { abbr: { text: string, bytes: integer, width: integer, hl_group: string }, kind: { text: string, bytes: integer, width: integer, hl_group: string }, menu: { text: string, bytes: integer, width: integer, hl_group: string } }
+---@return { abbr: { text: string, bytes: integer, width: integer, hl_group: string|table }, kind: { text: string, bytes: integer, width: integer, hl_group: string|table }, menu: { text: string, bytes: integer, width: integer, hl_group: string|table } }
 entry.get_view = function(self, suggest_offset, entries_buf)
   local item = self:get_vim_item(suggest_offset)
   return self.cache:ensure('get_view:' .. tostring(entries_buf), function()
--- a/lua/cmp/types/vim.lua
+++ b/lua/cmp/types/vim.lua
@@ -7,9 +7,9 @@
 ---@field public empty 1|nil
 ---@field public dup 1|nil
 ---@field public id any
----@field public abbr_hl_group string|nil
----@field public kind_hl_group string|nil
----@field public menu_hl_group string|nil
+---@field public abbr_hl_group string|table|nil
+---@field public kind_hl_group string|table|nil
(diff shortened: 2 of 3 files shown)


xzbdmw:
That’s much simpler in cmp side👍

hrsh7th:
thank you!

PR #1975, opened 2024-07-06
Use `keyword_pattern` for `is_symbol` check

joshbode:
Use `keyword_pattern` in place of `char.is_symbol` to improve matching (especially useful for improving for `cmp-path` and `cmp-cmdline` matches in the command-line).

Falls back to original behaviour of using `char.is_symbol` if no `keyword_pattern` is provided/specified for a source.

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -389,8 +389,9 @@ end
 ---Match line.
 ---@param input string
 ---@param matching_config cmp.MatchingConfig
+---@param keyword_pattern string
 ---@return { score: integer, matches: table[] }
-entry.match = function(self, input, matching_config)
+entry.match = function(self, input, matching_config, keyword_pattern)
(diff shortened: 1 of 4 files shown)


joshbode:
with this change (and `keyword_pattern = "[a-zA-Z0-9_./-]*"` for `path` source):
image

without the change, no matches

iguanacucumber:
Hi i'm trying to implement your PR with other PRs in [magazine.nvim](https://github.com/iguanacucumber) ( a fork to relieve burden from hrsh7th and a "beta" nvim-cmp ). I implemented Yioneko's PR which caches alot of things, and revamps matcher.lua. However when implementing your PR, is_symbol is nil. And i have no idea why it is nil so what do you think ?
Here's the changed file: https://github.com/iguanacucumber/magazine.nvim/blob/main/lua/cmp/matcher.lua#L105

PR #2006, opened 2024-08-01
fix: Add newer fallback method for tree-sitter scopes

pynappo:
A few hours ago, the [nvim-treesitter main branch removed `locals.get_scopes` for the newer `locals.get`](https://github.com/nvim-treesitter/nvim-treesitter/commit/0297cc6182fa7bf9608777c180172c35553d6a26#diff-eed1b34be50be29ccfe659070eb48f6ed610abaa7d984e3c5b8ac112cb1cfa38) so this PR adds a fallback to the newer method to help those trying out that branch.


hrsh7th:
thank you

Issue #820, opened 2022-03-01
Error executing lua: Vim:E976: using Blob as a String, Puppeteer completion


Abh15h3k:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp#faq) and it didn't resolve my problem.

### Issues

- [X] I have checked [existing issues](https://github.com/hrsh7th/nvim-cmp/issues) and there are no open or closed issues with the same problem.

### Neovim Version

NVIM v0.7.0-dev+1151-g9d3370a14

### Minimal reproducible config

same as provided minimal config with following addition

```
require'lspconfig'.tsserver.setup {
  capabilities = capabilities
}
```

### Description

neovim throws an error before giving suggestions.

### Steps to reproduce

open any `.js` or `.ts` file in a project where `puppeteer` is installed and start typing `page.keyboard.press('`

### Expected behavior

suggestions show without any error.

### Actual behavior

suggestions show after throwing the following error.

```
Error executing vim.schedule lua callback: ...are/nvim/site/pack/packer/opt/nvim-cmp/lua/cmp/entry.lua:201: Error executing lua: Vim:E976: using Blob as a String
stack traceback:
        [C]: in function 'strdisplaywidth'
        ...are/nvim/site/pack/packer/opt/nvim-cmp/lua/cmp/entry.lua:205: in function <...are/nvim/site/pack/packer/opt/nvim-cmp/lua/cmp/entry.lua:201>
        [C]: in function 'nvim_buf_call'
        ...are/nvim/site/pack/packer/opt/nvim-cmp/lua/cmp/entry.lua:201: in function 'callback'
        ...im/site/pack/packer/opt/nvim-cmp/lua/cmp/utils/cache.lua:36: in function 'get_view'
        ...packer/opt/nvim-cmp/lua/cmp/view/custom_entries_view.lua:122: in function 'open'
        ...hare/nvim/site/pack/packer/opt/nvim-cmp/lua/cmp/view.lua:110: in function 'open'
        ...hare/nvim/site/pack/packer/opt/nvim-cmp/lua/cmp/core.lua:328: in function 'fn'
        ...im/site/pack/packer/opt/nvim-cmp/lua/cmp/utils/async.lua:41: in function <...im/site/pack/packer/opt/nvim-cmp/lua/cmp/utils/async.lua:39>
stack traceback:
        [C]: in function 'nvim_buf_call'
[...]

PR #2050, opened 2024-09-27
fix: don't blow up when `nvim_buf_get_lines()` returns Blobs

wincent:
Some LSP servers may return binary garbage and `nvim_buf_get_lines()` will return a `Blob` instead of a `String` in those cases.

I added some `print(vim.inspect())` debugging in
`entry.get_documentation()` to prove that by the time the text passes through there, it's already garbage.

Here's an excerpt from a sample line returned by `nvim_buf_get_lines()`, as rendered by `vim.inspect()`:

    default\0\0\0!      vim.opt.background = 'dark'\0\0\0000

(etc)

Now, this looks like an LSP bug to me, but I think we shouldn't allow buggy LSP output to crash nvim-cmp. "Be conservative in what you send, be liberal in what you accept" and all that.

So, degrade by coercing any `Blob` we see into a `String` before passing it to `strdisplaywidth()`.

Closes: https://github.com/hrsh7th/nvim-cmp/issues/820

### DIFF
--- a/lua/cmp/utils/window.lua
+++ b/lua/cmp/utils/window.lua
@@ -309,6 +309,9 @@ window.get_content_height = function(self)
   local height = 0
   vim.api.nvim_buf_call(self:get_buffer(), function()
     for _, text in ipairs(vim.api.nvim_buf_get_lines(self:get_buffer(), 0, -1, false)) do
+      if vim.fn.type(text) == vim.v.t_blob then
+        text = vim.fn.string(text)
+      end
       height = height + math.max(1, math.ceil(vim.fn.strdisplaywidth(text) / self.style.width))
     end
   end)


hrsh7th:
Could you explain why and when `nvim_buf_get_lines` returns Blob instead of string...?

IMO, neovim's docs does not describe this behavior.

wincent:
> Could you explain why and when `nvim_buf_get_lines` returns Blob instead of string...?
> 
> IMO, neovim's docs does not describe this behavior.

Sadly, I have no idea why. I just know that it happens. I suspect that a buggy language server (in this case, [`lua-language-server`](https://github.com/LuaLS/lua-language-server)) is somehow producing the right kind of binary garbage — likely an invalid string encoding — to cause Neovim to encode it as a Blob. But I believe it's not limited to lua-language-server, based on [the comments on #820](https://github.com/hrsh7th/nvim-cmp/issues/820), which mention this happening in various scenarios.

Whatever the bug is, I'm seeing it when nvim-cmp shows the hover documentation for Enum types; note the artifacts in the text:

Enum type, showing the bug

For non-Enum types, it works fine:

non-Enum type, working fine

hrsh7th:
thank you.

PR #1980, opened 2024-07-09
perf: improve for source providing huge list of items

yioneko:
### Problem

nvim-cmp is still not ideally responsive when dealing with huge list of completion items, especially with tailwindcss which sends thousands of candidates on each completion response: #1009 #1828 #1949. Though some improvements have been done before but cmp is still apparently slower than other completion engine like vscode and coc.nvim in extreme cases.

### Attempt to improve

This PR extracts two commits from [my experimental perf branch fork](https://github.com/yioneko/nvim-cmp/tree/perf) which I think could be better upstreamed. The details and rationales are explained in commit descriptions.

The commits introduce massive changes to the codebase, and I'm not sure if you'd like to accept them. Since the changes are made just for improving performance in extreme cases, it's totally fine for me if you consider that it is not worth it.

### Comparison
Recorded at https://github.com/hrsh7th/nvim-cmp/commit/a110e12d0b58eefcf5b771f533fc2cf3050680ac:
cmp2

PR:
cmp1

PR #2060, opened 2024-10-11
perf: avoid source slowdown due to unnecessary context switches

jasonchoimtt:
Hello. I faced some completion performance issue, and I ended up tracking down a performance issue in `cmp/source.lua`.

The loop in question loops over all completion entries to perform filtering, so it is quite a tight loop. However, it yields to the event loop after every iteration. I think it results in too many context switches, slowing down the code.

Sorry that I don't have a good repro for this, by FWIW, in my own case making this change improves latency observably, when using the cmp-buffer source with a large number of words.

### DIFF
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -107,7 +107,7 @@ source.get_entries = function(self, ctx)
   ---@type cmp.Entry[]
   local entries = {}
   local matching_config = self:get_matching_config()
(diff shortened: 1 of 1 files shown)


iguanacucumber:
#1980 introduce more caching however it still yields after every iteration so i think that not yielding every time is a good thing

jasonchoimtt:
Thanks, appreciate the work!

PR #2069, opened 2024-10-21
fix: use unsimplified key for keymap.normalize

zeertzjq:
Fix https://github.com/hrsh7th/nvim-cmp/pull/1935#issuecomment-2426250159


hrsh7th:
thank you.

Issue #2070, opened 2024-10-23
<C-;> cmp mapping doesn't work


KaranveerB:
### FAQ

- [X] I have checked the [FAQ](https://github.com/hrsh7th/nvim-cmp/blob/main/doc/cmp.txt) and it didn't resolve my problem.

### Announcement

- [X] I have checked [Breaking change announcement](https://github.com/hrsh7th/nvim-cmp/issues/231).

### Minimal reproducible full config

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

let s:plug_dir = expand('/tmp/plugged/vim-plug')
if !filereadable(s:plug_dir .. '/plug.vim')
  execute printf('!curl -fLo %s/autoload/plug.vim --create-dirs https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim', s:plug_dir)
end

execute 'set runtimepath+=' . s:plug_dir
call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

" Setup global configuration. More on configuration below.
lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
      vim.fn["vsnip#anonymous"](args.body)
    end,
  },

  mapping = {
    ['<C-;>'] = cmp.mapping.confirm({ select = true }) -- only change
  },

  sources = cmp.config.sources({
    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}
EOF

lua << EOF
local capabilities = require('cmp_nvim_lsp').default_capabilities()

require'lspconfig'.cssls.setup {
  capabilities = capabilities,
}
EOF
```

### Description

`<C-;>` should confirm the action, but it does not. I'm not sure how many other key combos are affected.

On my regular config, I bisected to find the issue is from 2c93d5e. `<C-;>` seems to be a valid mapping for other neovim things so nvim-cmp should accept it too. 19bd8c7 doesn't fix this and the issue is present in the current latest commit 29fb485.

This may be an issue with nvim and its implementation of `keytrans`, but I'm busy and being lazy so I hope you'll forgive me for not looking too much more into it for now.

[...]

PR #2073, opened 2024-10-26
fix: make <c-;> work again

phanen:
Fix https://github.com/hrsh7th/nvim-cmp/issues/2070 https://github.com/hrsh7th/nvim-cmp/pull/1935#issuecomment-2442337135


hrsh7th:
Thank you.

PR #2081, opened 2024-11-04
fix: some symbols can't be mapped correctly

### LINKED ISSUE #2080: Can't add Command key(D-/) bindings in MacOS
That's strange because the mapping was working before. I don't know why it suddenly stopped working.
```lua
cmp.setup({
  mapping = {
    -- open / close
    ["<D-/>"] = cmp.mapping({
      i = function()
        if cmp.visible() then
          cmp.abort()
        else
          cmp.complete()
        end
      end,
      c = function()
        if cmp.visible() then
          cmp.close()
        else
          cmp.complete()
        end
      end,
    }),
  },
})
```

OnlyWick:
#1935 will cause some symbols to not be mapped correctly. I have added the symbols that I need to map. I'm not very familiar with others' habits, but I believe there should be some other symbols that need to be added to this patter.

Fix #2080.

### DIFF
--- a/lua/cmp/utils/keymap.lua
+++ b/lua/cmp/utils/keymap.lua
@@ -8,7 +8,7 @@ local keymap = {}
 ---@param keys string
 ---@return string
 keymap.t = function(keys)
-  return (string.gsub(keys, '(<[A-Za-z0-9\\%-%[%]%^@;_]->)', function(match)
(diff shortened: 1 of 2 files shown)


EasonMo (approved):
Passed the test with <M+/> and can run correctly.

fell-z (approved):
My keymap `<C-.>` worked correctly.

hrsh7th:
@OnlyWick Could you resolve conflicts?

OnlyWick:
@hrsh7th Done!

PR #2087, opened 2024-11-09
fix: make sure nvim-cmp keymaps are registered

tronikelis:
## Context

When editing files with the https://github.com/microsoft/compose-language-service LSP, the popup would always appear on new lines like so:

<details>
<summary>Completion appears on fully whitespace line</summary>

As you can see, this interferes with my fallback Tab keymap which indents the code and its very annoying

https://github.com/user-attachments/assets/865dcb18-72f8-4d67-8b07-39e29aaa36db

This is with `keyword_length = 1`

</details>

## The underlying issue

I think the underlying issue is:
1. docker compose LSP acts differently (this does not happen on other lsps)
2. nvim-cmp, even with `keyword_length = 1` still shows the menu when there are no keywords (fully whitespace line), unless this is the expected behavior

## Workaround

I started using this config to workaround the issue

```
enabled = function()
    return #vim.trim(vim.api.nvim_get_current_line()) ~= 0
end,
```

But then I saw that sometimes my completion selection just stopped working altogether like so: 

<details>
<summary>Completion stops</summary>

https://github.com/user-attachments/assets/b8436986-e773-484f-9265-ebbf0301dbc0

</details>

## This PR

This pr makes the workaround actually work

<details>
<summary>With this PR</summary>

https://github.com/user-attachments/assets/70e812ae-def6-4e6d-bdc1-1755cb4c3168

</details>

### DIFF
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -328,9 +328,13 @@ cmp.setup = setmetatable({
   end,
 })
 
+local cmp_prepared = false
+
 -- In InsertEnter autocmd, vim will detects mode=normal unexpectedly.
 local on_insert_enter = function()
-  if config.enabled() then
+  if not cmp_prepared or config.enabled() then
+    cmp_prepared = true
+
     cmp.config.compare.scopes:update()
     cmp.config.compare.locality:update()
     cmp.core:prepare()
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -334,10 +334,12 @@ local cmp_prepared = false
 local on_insert_enter = function()
   if not cmp_prepared or config.enabled() then
     cmp_prepared = true
-
     cmp.config.compare.scopes:update()
     cmp.config.compare.locality:update()
     cmp.core:prepare()
+  end
+
+  if config.enabled() then
     cmp.core:on_change('InsertEnter')
   end
 end


hrsh7th:
IMO, you can use `get_trigger_characters` option.

check out it.

```lua
require('cmp').setup({
  ...
  completion = {
    get_trigger_characters = function(trigger_characters)
      return vim.iter(trigger_characters):filter(function(char)
        return not chat:match('^%s*$')
      end)
    end
  }
  ...
})
```

tronikelis:
> IMO, you can use `get_trigger_characters` option.
> 
> check out it.
> 
> ```lua
> require('cmp').setup({
>   ...
>   completion = {
>     get_trigger_characters = function(trigger_characters)
>       return vim.iter(trigger_characters):filter(function(char)
>         return not chat:match('^%s*$')
>       end)
>     end
>   }
>   ...
> })
> ```

Damn, nice, seems to work, but I am not finding this in the doc?

tronikelis:
With your config option there is a problem,

When completing a function the signature help and completion window does not open by itself anymore (which I want)

tronikelis:
Do you see any problems with this fix?

tronikelis:
this seems to work

```lua
get_trigger_characters = function(trigger_characters)
    if #vim.trim(vim.api.nvim_get_current_line()) == 0 then
        return {}
    end

    return trigger_characters
end,
```

nmrtv:
This commit introduced a bug where the 'enabled' option is not respected in prompt buffers. For example, in the Telescope prompt, it provides autocomplete, which is quite annoying.

tronikelis:
> This commit introduced a bug where the 'enabled' option is not respected in prompt buffers. For example, in the Telescope prompt, it provides autocomplete, which is quite annoying.

hmm i cant seem to reproduce, what is your config?

nmrtv:
The configuration is fairly standard. With this commit enabled, you can open Telescope's live grep, type a few characters, and try navigating the search results using the standard Ctrl+n and Ctrl+p shortcuts. However, instead of navigating through the Telescope search results, the nvim-cmp completion menu appears.

nmrtv:
This happens because nvim-cmp is disabled in prompt buffers. However, with this commit, it is enabled in prompts.

tronikelis:
> This happens because nvim-cmp is disabled in prompt buffers. However, with this commit, it is enabled in prompts.

doesnt happen in my config which I havent touched the `enabled`, can you send your `enabled` function?

nmrtv:
It's default: https://github.com/hrsh7th/nvim-cmp/blob/ed31156aa2cc14e3bc066c59357cc91536a2bc01/lua/cmp/config/default.lua#L10

nmrtv:
I don't know if it helps, but nvim-cmp mappings are the same as telescope:
```lua
			["<C-n>"] = cmp.mapping(function(fallback)
				if cmp.visible() then
					cmp.select_next_item()
					cmp.complete()
				else
					fallback()
				end
			end, { "i", "c", "s" }),
			["<C-p>"] = cmp.mapping(function(fallback)
				if cmp.visible() then
					cmp.select_prev_item()
				else
					fallback()
				end
			end, { "i", "c", "s" }),
```

tronikelis:
> I don't know if it helps, but nvim-cmp mappings are the same as telescope:
> 
> ```lua
> 			["<C-n>"] = cmp.mapping(function(fallback)
> 				if cmp.visible() then
> 					cmp.select_next_item()
> 				elseif has_words_before() then
> 					cmp.complete()
> 				else
> 					fallback()
> 				end
> 			end, { "i", "c", "s" }),
> 			["<C-p>"] = cmp.mapping(function(fallback)
> 				if cmp.visible() then
> 					cmp.select_prev_item()
> 				else
> 					fallback()
> 				end
> 			end, { "i", "c", "s" }),
> ```

thanks, however it shouldn't matter based on your keymaps, maybe I misunderstood what cmp:prepare does, 

I just pushed a quick commit that tries to call less stuff, can you check it? If it still does not work then I will just use the alternative which is in an earlier comment

nmrtv:
> thanks, however it shouldn't matter based on your keymaps, maybe I misunderstood what cmp:prepare does,

Well, I think it matters because this issue wouldn’t occur if the mappings in Telescope and nvim-cmp were different, and ctrl+n and ctrl+p didn’t trigger nvim-cmp completion. Thanks, I'll look at your fix tomorrow.

tronikelis:
> > thanks, however it shouldn't matter based on your keymaps, maybe I misunderstood what cmp:prepare does,
> 
> Well, I think it matters because this issue wouldn’t occur if the mappings in Telescope and nvim-cmp were different, and ctrl+n and ctrl+p didn’t trigger nvim-cmp completion. Thanks, I'll look at your fix tomorrow.

I have the same c-n c-p mappings

tronikelis:
closing in favor of https://github.com/hrsh7th/nvim-cmp/pull/2087#issuecomment-2491503971

PR #2107, opened 2024-12-06
fix: make <M-`> work again

danielpclin:
Identical fix to https://github.com/hrsh7th/nvim-cmp/pull/2073

### DIFF
--- a/lua/cmp/utils/keymap_spec.lua
+++ b/lua/cmp/utils/keymap_spec.lua
@@ -23,6 +23,7 @@ describe('keymap', function()
       '<C-/>',
       '<C-_>',
       "<C-'>",
+      'M-`',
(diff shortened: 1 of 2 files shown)


lua/cmp/utils/keymap_spec.lua
```diff
@@ -23,6 +23,7 @@ describe('keymap', function()
       '<C-/>',
       '<C-_>',
       "<C-'>",
+      'M-`',
```
hrsh7th, line 26: This should be `<M-\`>` instead?

danielpclin: Yes, I meant "<M-`>"

hrsh7th:
Thank you.

PR #2117, opened 2024-12-20
Fix :checkhealth warnings

jdtoombs:
Fix the following warning when running `:checkhealth`

  ```
  - WARNING vim.str_utfindex is deprecated. Feature will be removed in Nvim 1.0
  - ADVICE:
    - use vim.str_utfindex(s, encoding, index, strict_indexing) instead.
  ```
```typescript
function vim.str_utfindex(s: string, encoding: "utf-16"|"utf-32"|"utf-8", index?: integer, strict_indexing?: boolean)
  -> integer
```

Additionally, update vim.str_byteindex to match:
```typescript
function vim.str_byteindex(s: string, encoding: "utf-16"|"utf-32"|"utf-8", index: integer, strict_indexing?: boolean)
  -> integer
```

### DIFF
--- a/lua/cmp/utils/misc.lua
+++ b/lua/cmp/utils/misc.lua
@@ -197,7 +197,7 @@ end
 ---@return integer
 misc.to_utfindex = function(text, vimindex)
   vimindex = vimindex or #text + 1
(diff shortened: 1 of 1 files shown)


gepbird:
This was done in c4c784a6

PR #2121, opened 2025-01-05
Mark vimscript block as "vim"

WhyNotHugo:
The example configuration is Vimscript, with a heredoc containing some Lua code. Using "lua" as a language identifier results neither the Vimscript portion nor the Lua portion being highlighted propertly.

Mark the code block as "vim".  Neovim (with treesitter) properly highlights the outer block as Vimscript, and the Lua heredoc as Lua code.


hrsh7th:
Thank you.

PR #2126, opened 2025-01-18
fix: Outdated completion item with mini.snippets

### LINKED ISSUE #2119: Outdated completion item with mini.snippets
### FAQ

### Announcement

repro.lua, run with `nvim -u repro.lua`
```lua
local root = vim.fn.fnamemodify("./.repro", ":p")
-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
    vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end
-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        "git",
        "clone",
        "--filter=blob:none",
        "--single-branch",
        "https://github.com/folke/lazy.nvim.git",
        lazypath,
    })
end
vim.opt.runtimepath:prepend(lazypath)
-- install plugins
local plugins = {
    -- do not remove the colorscheme!
    "folke/tokyonight.nvim",
    "neovim/nvim-lspconfig",
    {
        "hrsh7th/nvim-cmp",
        lazy = false,
        dependencies = {
            "hrsh7th/cmp-nvim-lsp",
        },
        config = function(_, opts)
            local cmp = require("cmp")
            require("cmp").setup({
                mapping = cmp.mapping.preset.insert({
                    ["<cr>"] = cmp.mapping(function(fallback)
                        if cmp.visible() then
                            cmp.confirm()
                        end
                    end, { "i", "c", "s" }),
                }),
                completion = {
                    completeopt = "menu,menuone,noinsert",
                },
                snippet = {
                    expand = function(args)
                        local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
                        insert({ body = args.body })
                    end,
                },
                sources = require("cmp").config.sources({
                    { name = "nvim_lsp" },
                }, {}),
            })
        end,
    },
    {
        "echasnovski/mini.snippets",
        version = false,
        config = function()
            local gen_loader = require("mini.snippets").gen_loader
            require("mini.snippets").setup({
                snippets = {
                    -- Load custom file with global snippets first (adjust for Windows)
                    gen_loader.from_file("~/.config/nvim/snippets/global.json"),

                    -- Load snippets based on current language by reading files from
                    -- "snippets/" subdirectories from 'runtimepath' directories.
                    gen_loader.from_lang(),
                },
            })
        end,
    },
    -- add any other pugins here
}

require("lazy").setup(plugins, {
    root = root .. "/plugins",
})

require("lspconfig").lua_ls.setup({
    settings = {
        capabilities = require("cmp_nvim_lsp").default_capabilities(),
        Lua = {
            runtime = {
                version = "LuaJIT",
            },
            workspace = {
                library = {
                    "/usr/local/share/nvim/runtime",
                },
            },
            completion = {
                callSnippet = "Replace",
            },
        },
    },
})
vim.cmd([[colorscheme tokyonight]])
```

### Description

Some chars are swallowed when expanding:
typing sequence : `vim.sc<cr>fun<cr>` result in right pair missing;
typing sequence : `vim.sc<cr><cr>` has correct result,
typing sequence : `vim.schedule()<left>fun<cr>` has correct result,

https://github.com/user-attachments/assets/8a8d353e-27f1-4c1e-950e-8e1650a4bbde

I debug to find cmp removing the pair before snippet expanding, needs more investment.

The whole process is following, `|` stands for cursor:
1. type `vim.sc<cr>` expanding to `vim.schedule(|fn)`
2. lua_ls interprets it as if user types `vim.schedule(fn)` suggesting `fun` as a snippet completion item
3. user continues typing  `fun`, completion item stays unchanged, but cmp knows before expanding the snippets, it needs to clear the `fun` typed by user, to let snippet engine expand correctly.
4. user types `<cr>`, cmp turn `vim.schedule(fun)` to `vim.schedule()` correctly, **BUT**, it calls [vim.lsp.util.apply_text_edits](https://github.com/hrsh7th/nvim-cmp/blob/b555203ce4bd7ff6192e759af3362f9d217e8c89/lua/cmp/core.lua#L489) before [expanding](https://github.com/hrsh7th/nvim-cmp/blob/b555203ce4bd7ff6192e759af3362f9d217e8c89/lua/cmp/core.lua#L497-L500), while the TextEdit looks like this:
```txt
cmp stage before { "vim.schedule()" } -- This is before apply text edit
function#function#if completion_item.textEdit: {
  _index = 1,
  newText = "",
  range = {
    ["end"] = {
      character = 15, 
      line = 765
    },
    start = {
      character = 13,
      line = 765
    }
  }
}
cmp stage after { "vim.schedule(" } -- This is after appying
```
notice  the range is [13,15) which stands for `fn` in the original placeholder.

`vim.snippet.expand` works correctly because it stays in select mode, does not trigger completion at all, only after user typing `fun` lua_ls will begin to send correct TextEdits, anyway, by that time placeholder is already gone.

To my understanding, the fix would be suspending completion request to stop step 2 from firing when first expanding.

### Steps to reproduce

above

### Expected behavior

no outdated completion item

### Actual behavior

outdated completion item swallows ')'

### Additional context

_No response_

abeldekat:
Fixes #2119 

The issue as described by @xzbdmw has been debated in several other places:

[.. triggers completion when it should not ..](https://github.com/abeldekat/cmp-mini-snippets/issues/6)
[beta testing mini.snippets: expanding snippets from lsp](https://github.com/abeldekat/cmp-mini-snippets/issues/9)

@xzbdmw provided a clear explanation:

>>I'll try to explain more detail:
mini.snippets: insert mode -> type 'f' -> both mini.snippets and cmp detects TextChangedI -> cmp-nvim-lsp sends request -> lua_ls receives the buffer content, which is vim.schedule(ffn) -> mini.snippets clears the placeholder fn, it becomes vim.schedule(f) -> "fun" completion_item contains TextEdits that want to clear the fn , now what in fn's original place is now the right pair -> right pair gone, before snippets expanding. Neither mini.snippets or cmp is to blame here IMO, they all do what they supposed to do.

>>luasnip: select mode -> type 'f' -> nvim clears the placeholder, by the definition of select mode -> cmp-nvim-lsp sends request -> lua_ls receives the buffer content, which is vim.schedule(f) -> all good.

@hrsh7th

I considered solving the problem in `cmp-nvim-lsp`, introducing some kind of flag indicating whether the call to `complete` should be scheduled or not.

Then I saw that `TextChangedI` and `TextChangedP` do not yet use `async.debounce_next_tick`. 

I think this solves the issue on a more appropriate level.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,10 @@ Plug 'hrsh7th/vim-vsnip'
 " Plug 'L3MON4D3/LuaSnip'
 " Plug 'saadparwaiz1/cmp_luasnip'
 
+" For mini.snippets users.
+" Plug 'echasnovski/mini.snippets'
+" Plug 'abeldekat/cmp-mini-snippets'
+
 " For ultisnips users.
 " Plug 'SirVer/ultisnips'
 " Plug 'quangnguyen30192/cmp-nvim-ultisnips'
@@ -69,6 +73,12 @@ lua <<EOF
         -- require('snippy').expand_snippet(args.body) -- For `snippy` users.
         -- vim.fn["UltiSnips#Anon"](args.body) -- For `ultisnips` users.
         -- vim.snippet.expand(args.body) -- For native neovim snippets (Neovim v0.10+)
+
+        -- For `mini.snippets` users:
+        -- local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
+        -- insert({ body = args.body }) -- Insert at cursor
+        -- cmp.resubscribe({ "TextChangedI", "TextChangedP" })
+        -- require("cmp.config").set_onetime({ sources = {} })
       end,
     },
     window = {
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -62,6 +62,10 @@ A recommended configuration can be found below.
   " Plug 'L3MON4D3/LuaSnip'
   " Plug 'saadparwaiz1/cmp_luasnip'
 
+  " For mini.snippets users.
+  " Plug 'echasnovski/mini.snippets'
+  " Plug 'abeldekat/cmp-mini-snippets'
+
   " For snippy users.
   " Plug 'dcampos/nvim-snippy'
   " Plug 'dcampos/cmp-snippy'
@@ -86,6 +90,12 @@ A recommended configuration can be found below.
           -- require'snippy'.expand_snippet(args.body) -- For `snippy` users.
           -- vim.fn["UltiSnips#Anon"](args.body) -- For `ultisnips` users.
           -- vim.snippet.expand(args.body) -- For native neovim snippets (Neovim v0.10+)
+
+          -- For `mini.snippets` users:
+          -- local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
+          -- insert({ body = args.body }) -- Insert at cursor
+          -- cmp.resubscribe({ "TextChangedI", "TextChangedP" })
+          -- require("cmp.config").set_onetime({ sources = {} })
         end,
       },
       window = {
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -494,7 +494,6 @@ core.confirm = function(self, e, option, callback)
         (#texts == 1 and (completion_item.textEdit.range.start.character + #texts[1]) or #texts[#texts]),
       })
       if is_snippet then
-        config.set_onetime({ sources = {} })
         config.get().snippet.expand({
           body = new_text,
           insert_text_mode = completion_item.insertTextMode,
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -494,6 +494,7 @@ core.confirm = function(self, e, option, callback)
         (#texts == 1 and (completion_item.textEdit.range.start.character + #texts[1]) or #texts[#texts]),
       })
       if is_snippet then
+        config.set_onetime({ sources = {} })
         config.get().snippet.expand({
           body = new_text,
           insert_text_mode = completion_item.insertTextMode,
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -328,6 +328,12 @@ cmp.status = function()
   end
 end
 
+---Ensures that cmp is the last receiver of the events specified.
+---@param events string[]
+cmp.resubscribe = function(events)
+  autocmd.resubscribe(events)
+end
+
 ---@type cmp.Setup
 cmp.setup = setmetatable({
   global = function(c)
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -366,7 +366,7 @@ local on_text_changed = function()
     cmp.core:on_change('TextChanged')
   end
 end
-autocmd.subscribe({ 'TextChangedI', 'TextChangedP' }, on_text_changed)
+autocmd.subscribe({ 'TextChangedI', 'TextChangedP' }, async.debounce_next_tick(on_text_changed))
 autocmd.subscribe('CmdlineChanged', async.debounce_next_tick(on_text_changed))
 
 autocmd.subscribe('CursorMovedI', function()
--- a/lua/cmp/init.lua
+++ b/lua/cmp/init.lua
@@ -366,7 +366,7 @@ local on_text_changed = function()
     cmp.core:on_change('TextChanged')
   end
 end
-autocmd.subscribe({ 'TextChangedI', 'TextChangedP' }, async.debounce_next_tick(on_text_changed))
+autocmd.subscribe({ 'TextChangedI', 'TextChangedP' }, on_text_changed)
 autocmd.subscribe('CmdlineChanged', async.debounce_next_tick(on_text_changed))
 
 autocmd.subscribe('CursorMovedI', function()
--- a/lua/cmp/utils/autocmd.lua
+++ b/lua/cmp/utils/autocmd.lua
@@ -6,6 +6,16 @@ autocmd.group = vim.api.nvim_create_augroup('___cmp___', { clear = true })
 
 autocmd.events = {}
 
+local function create_autocmd(event)
+  vim.api.nvim_create_autocmd(event, {
+    desc = ('nvim-cmp: autocmd: %s'):format(event),
+    group = autocmd.group,
+    callback = function()
+      autocmd.emit(event)
+    end,
+  })
+end
+
 ---Subscribe autocmd
 ---@param events string|string[]
 ---@param callback function
@@ -16,13 +26,7 @@ autocmd.subscribe = function(events, callback)
   for _, event in ipairs(events) do
     if not autocmd.events[event] then
       autocmd.events[event] = {}
-      vim.api.nvim_create_autocmd(event, {
-        desc = ('nvim-cmp: autocmd: %s'):format(event),
-        group = autocmd.group,
-        callback = function()
-          autocmd.emit(event)
-        end,
-      })
+      create_autocmd(event)
     end
     table.insert(autocmd.events[event], callback)
   end
@@ -47,6 +51,26 @@ autocmd.emit = function(event)
   autocmd.events[event] = autocmd.events[event] or {}
   for _, callback in ipairs(autocmd.events[event]) do
     callback()
+  end
+end
+
+---Resubscribe to events
+---@param events string[]
+autocmd.resubscribe = function(events)
+  -- Delete the autocommands if present
+  local found = vim.api.nvim_get_autocmds({
+    group = autocmd.group,
+    event = events,
+  })
+  for _, to_delete in ipairs(found) do
+    vim.api.nvim_del_autocmd(to_delete.id)
+  end
+
+  -- Recreate if event is known
+  for _, event in ipairs(events) do
+    if autocmd.events[event] then
+      create_autocmd(event)
+    end
   end
 end
 


abeldekat:
It is possible to solve the problem in the user's [config](https://github.com/abeldekat/cmp-mini-snippets/issues/9#issuecomment-2586287824):


```lua
local function add_cmp()
  local group = vim.api.nvim_create_augroup("mini_snippets_nvim_cmp", { clear = true })

  -- NOTE: Firstly, make sure that nvim-cmp never provides completion items directly after snippet insert
  -- See https://github.com/abeldekat/cmp-mini-snippets/issues/6
  vim.api.nvim_create_autocmd("User", {
    group = group,
    pattern = "MiniSnippetsSessionStart",
    callback = function(_) require("cmp.config").set_onetime({ sources = {} }) end,
  })

  -- HACK: Secondly, it's now possible to prevent outdated completion items
  -- See https://github.com/hrsh7th/nvim-cmp/issues/2119
  local function make_complete_override(complete_fn)
    return function(self, params, callback)
      local override_fn = complete_fn
      if MiniSnippets.session.get(false) ~= nil then override_fn = vim.schedule_wrap(override_fn) end
      override_fn(self, params, callback)
    end
  end
  local function find_cmp_nvim_lsp(id)
    for _, source in ipairs(require("cmp").get_registered_sources()) do
      if source.id == id and source.name == "nvim_lsp" then return source.source end
    end
  end
  vim.api.nvim_create_autocmd("User", {
    group = group,
    pattern = "CmpRegisterSource",
    callback = function(ev)
      local cmp_nvim_lsp = find_cmp_nvim_lsp(ev.data.source_id)
      if cmp_nvim_lsp then
        local org_complete = cmp_nvim_lsp.complete
        cmp_nvim_lsp.complete = make_complete_override(org_complete)
      end
    end,
  })

  MiniDeps.add({ source = "hrsh7th/nvim-cmp", depends = { "hrsh7th/cmp-nvim-lsp", "hrsh7th/cmp-buffer" } })
  local cmp = require("cmp")
  require("cmp").setup({
    snippet = {
      expand = function(args) -- use mini.snippets to expand snippets from lsp
        local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
        insert({ body = args.body }) -- Insert at cursor
      end,
    },
    sources = cmp.config.sources({ { name = "nvim_lsp" }, { name = "buffer" } }),
    mapping = cmp.mapping.preset.insert(),
    completion = { completeopt = "menu,menuone,noinsert" },
  })
end


```

hrsh7th:
In my career as a front-end developer, solving any kind of asynchronous timing problem with things like setTimeout or vim.schedule has always been an anti-pattern.

I think this problem should be solved by mini.snippets.

echasnovski:
> I think this problem should be solved by mini.snippets.

As far as I understand, there is no problem in 'mini.snippets' to solve. The way it works mostly served as a source to reproduce an issue in completion plugins, which is roughly as follows: there can be other text changes in between completion plugin reacting to 'TextChangedI' / 'TextChangedP' and async action they do as a result. This might lead to action being outdated as the buffer text state has already changed.

The 'mini.snippets' showed this issue because it can react on text change by removing tabstop placeholder *immediately* in the same tick (otherwise there will be flicker) inside 'TextChanged{I,P}' callback. Other snippet engines utilize Select mode here: its replacing effect is done internally by Neovim and is already in place for 'TextChanged{I,P}' events.

hrsh7th:
I didn't know about that behavior of mini.snippets.

I'm guessing that mini.snippets changes the text during TextChangedI/P, which then causes TextChangedI/P to fire again and complete correctly, but let's look into it.

echasnovski:
> I'm guessing that mini.snippets changes the text during TextChangedI/P, which then causes TextChangedI/P to fire again and complete correctly, but let's look into it.

Yes, removing placeholder in 'mini.snippets' indeed triggers either `TextChangedI` or `TextChangedP` (after ones caused by the user typing a character which causes placeholder removal). They are not suppressed because there is a text change in Insert mode which those events are meant monitor.

hrsh7th:
I found that this doesn't happen with the new completion engine I'm developing, but it does happen with nvim-cmp.

So this is something nvim-cmp wants to solve somehow.

But it seems like it's going to be pretty difficult to solve...

hrsh7th:
nvim-cmp has a lot of problems...

Plus, rivals have emerged, so I'm motivated.

- views are pluggable
- macro support (there are asynchronous issues in some cases, but it works in most cases)
- tests have been written

etc.

It might be a good idea to release it as a beta test, but I'm afraid the interface will freeze.

hrsh7th:
@abeldekat Could you try this?

abeldekat:
@hrsh7th,

Sure!

<details>
<summary>minimal init</summary>

```lua
-- Clone latest 'mini.nvim' (requires Git CLI installed)
vim.cmd('echo "Installing `mini.nvim`" | redraw')
local mini_path = vim.fn.stdpath("data") .. "/site/pack/deps/start/mini.nvim"
local clone_cmd = { "git", "clone", "--depth=1", "https://github.com/echasnovski/mini.nvim", mini_path }
vim.fn.system(clone_cmd)
vim.cmd('echo "`mini.nvim` is installed" | redraw')

-- Make sure 'mini.nvim' is available
vim.cmd("packadd mini.nvim")
require("mini.deps").setup()

-- Add extra setup steps needed to reproduce the behavior
-- Use `MiniDeps.add('user/repo')` to install another plugin from GitHub

require("mini.notify").setup() -- lsp loading indicator
local mini_notify = MiniNotify.make_notify() -- opts
---@diagnostic disable-next-line: duplicate-set-field
vim.notify = function(msg, level, _) mini_notify(msg, level) end

local mini_snippets = require("mini.snippets")
mini_snippets.setup({
  snippets = { -- just one dummy custom snippet:
    { prefix = "dummy", body = "T1=fu$1", desc = "fu before $1" },
  },
})

MiniDeps.add({ source = "hrsh7th/nvim-cmp", depends = { "hrsh7th/cmp-nvim-lsp", "hrsh7th/cmp-buffer" } })
local cmp = require("cmp")
require("cmp").setup({
  snippet = {
    expand = function(args) -- use mini.snippets to expand snippets from lsp
      local insert = MiniSnippets.config.expand.insert or MiniSnippets.default_insert
      insert({ body = args.body }) -- Insert at cursor
    end,
  },
  sources = cmp.config.sources({ { name = "nvim_lsp" }, { name = "buffer" } }),
  mapping = cmp.mapping.preset.insert(),
  view = { docs = { auto_open = false } },
})

MiniDeps.add("williamboman/mason.nvim")
MiniDeps.add("williamboman/mason-lspconfig.nvim")
MiniDeps.add("neovim/nvim-lspconfig")
require("mason").setup()
require("mason-lspconfig").setup({ ensure_installed = { "lua_ls" } })

local capabilities = vim.lsp.protocol.make_client_capabilities()
capabilities = vim.tbl_deep_extend("force", capabilities, require("cmp_nvim_lsp").default_capabilities() or {})
require("lspconfig").lua_ls.setup({
  capabilities = capabilities,
  settings = {
    Lua = {
      runtime = { version = "LuaJIT" },
      completion = { callSnippet = "Replace", keywordSnippet = "Replace" },
      workspace = {
        checkThirdParty = false,
        library = { vim.env.VIMRUNTIME },
      },
    },
  },
})
```
</details>

<details>
<summary>test file</summary>

```lua
--[[
TLDR: functions with parameters annotated with `fun()`
fail to expand correctly when the user choses the `fun` completion item from lua_ls
Happens with both blink and nvim-cmp

-- NOTE: Mini.snippets is setup to only provide lsp expansion. No completion sources.

-- NOTE: Completion keys: `<c-y> <c-n> <c-p>``

-- NOTE: In all "wrong" cases, if `callback` is not nested(press <c-c>), expansion is correct!
  Conclusion: the lsp snippet `{body = "function ()\n\t$0\nend}"`, having kind "function", is not wrong itself

-- NOTE: After initial snippet expand no completions should be suggested.
  However, `nvim-cmp` does suggest completion items, but **only** after
  expansion of `demo4` and `demo5`, where callback is the first parameter.
  
--]]

-- cmp: correct
-- blink: correct
---@param message string
---@param callback function
local function demo1(message, callback) end

-- cmp: correct
-- blink: correct
---@param callback function
---@param message string
local function demo2(callback, message) end

-- cmp: select "fun" completion item -> wrong
-- blink: select "fun" completion item -> wrong
---@param message string
---@param callback fun() -- Two lsp snippet completions: fun and function
local function demo3(message, callback) end

-- cmp: select "fun" completion item -> correct!
-- blink: select "fun" completion item -> wrong
---@param callback fun() -- Two lsp snippet completions: fun and function
---@param message string
local function demo4(callback, message) end

-- cmp: select "fun" completion item -> wrong
-- blink: select "fun" completion item -> wrong
---@param callback fun() -- Two lsp snippet completions: fun and function
local function demo5(callback) end
```
</details>

<details>
<summary>video result</summary>

https://github.com/user-attachments/assets/4f0fd2ef-a6be-4730-ade4-4c980df64991

</details>

Unfortunately, the issue is still there. 

I would like to hear your thoughts on the [idea](https://github.com/hrsh7th/nvim-cmp/pull/2126#issuecomment-2601984174) I wrote down in my last comment.
As "pro" I can see:

1. No fixing code in `nvim-cmp`. Just a new function, `resubscribe`
2. No fixing code in `mini` either.
3. No vim.schedule. I do agree that this is an anti pattern.

As "con"

1. One extra line in the user's config: The call to `cmp.resubscribe`.

hrsh7th:
There would be no problem if nvim-cmp were simply equipped with resubscribe. I would accept it.

However, one thing that concerns me is that it may not be a fundamental fix, but that's a topic for another discussion.

(my words almost translated by google 😅

abeldekat:
@hrsh7th,

I added two new commits. It's a very concrete fix, not fundamental indeed.

@echasnovski 

Could you also have a look? I would like to conditionally call `cmp.resubscribe`, seeing  [this](https://github.com/echasnovski/mini.nvim/blob/fcf982a66df4c9e7ebb31a6a01c604caee2cd488/lua/mini/snippets.lua#L2092) code. Using `MiniSnippets.session.get(true)` does a full clone and is perhaps too expensive.

echasnovski:
> Using `MiniSnippets.session.get(true)` does a full clone and is perhaps too expensive.

On a typical use case when there is no snippet session it will just by `vim.deepcopy(nil)`, which is *very* fast. Even with active session, it is typically on the order of 0.01ms.

abeldekat:
> On a typical use case when there is no snippet session it will just by vim.deepcopy(nil), which is very fast. Even with active session, it is typically on the order of 0.01ms.

Thanks. I think it's not even necessary. I benchmarked the call to `cmp.resubscribe`, as you explained in [this](https://github.com/abeldekat/cmp-mini-snippets/issues/2) issue.

The `median` is `0.0475465` on an old laptop. I think that's not too bad, even when added after every expand.

hrsh7th:
Thank you.

abeldekat:
My pleasure!

PR #2141, opened 2025-03-13
Preserve 'buflisted' state when applying LSP text edits

ColdSpirit0:
fixes https://github.com/olimorris/codecompanion.nvim/issues/1060

This change preserves the original `buflisted` state of the current buffer when applying LSP text edits. The function `vim.lsp.util.apply_text_edits` forces `buflisted` to true (see https://github.com/neovim/neovim/issues/12488 and https://github.com/neovim/neovim/pull/12489), which is useful for non-current buffers but not desired for the current one. This fix saves the state before applying edits and restores it afterward.

Same as https://github.com/Saghen/blink.cmp/pull/1432

`make test` and `make lint` passed locally.

### DIFF
--- a/lua/cmp/core.lua
+++ b/lua/cmp/core.lua
@@ -433,11 +433,11 @@ core.confirm = function(self, e, option, callback)
           return
         end
         vim.cmd([[silent! undojoin]])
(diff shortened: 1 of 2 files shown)


hrsh7th:
IMO, this should be implemented in core. but I can merge this. thank you.

PR #2153, opened 2025-03-31
fix thumb weird look after using `vim.o.winborder = 'rounded'`

Coosis:
if using `vim.o.winborder = 'rounded'`, the scrollbar/scroll thumb is weird looking:
before

after fix:
after

this also needed to be fixed upstream, which had been merged recently: [pr](https://github.com/neovim/neovim/pull/32984)

As of right now, the neovim in `nixpkgs` is not yet fixed, have to use newest `homebrew` [neovim](https://github.com/Homebrew/homebrew-core/blob/dae846ae5347d31b610aebc2df6267f9a734f5a4/Formula/n/neovim.rb) and apply this fix to make it look normal


Coosis:
sorry didn't see the other pr

PR #2158, opened 2025-04-01
fix: Use winborder for window menu and fix scrollbar window

Aqothy:
Use win border as the default border option in config, also fix scrollbar window

scrollbar before: 
Screenshot 2025-04-01 at 4 43 10 PM

scrollbar after (also with rounded win border): 
Screenshot 2025-04-01 at 5 23 53 PM

### DIFF
--- a/lua/cmp/config/default.lua
+++ b/lua/cmp/config/default.lua
@@ -1,5 +1,6 @@
 local compare = require('cmp.config.compare')
 local types = require('cmp.types')
+local window = require('cmp.config.window')
 
 local WIDE_HEIGHT = 40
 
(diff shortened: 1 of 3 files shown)


Aqothy:
my first open contribution, would love if this could be merged :)

wsdjeg:
same as #2150

Aqothy:
> same as #2150

hey this pr also adds the feature to use the win border option as the default border option in cmp config, so if user has vim.opt.winborder = "rounded" they dont have to configure nvim cmp to use rounded borders using cmp.config.window.bordered(), up to you if you want this to be the default though

Issue #1649, opened 2023-07-08
Incorrect highlighting


HUAHUAI23:
When using nvim --headless and connecting with neovide, the highlighting of the cmp plugin is problematic, has anyone encountered this? 
<img width="892" alt="image" src="https://github.com/hrsh7th/nvim-cmp/assets/43649186/337058c9-332c-4718-8001-d871034d105c">

williamhCode:
I am working on a neovim gui and have encountered the same problem. https://github.com/neovim/neovim/issues/28120#issue-2216743114

zeertzjq:
This happens because 'termguicolors' is not set when nvim-cmp calls `synIDattr()`. It can be solved by setting 'termguicolors' before loading nvim-cmp.

zeertzjq:
> This happens because 'termguicolors' is not set when nvim-cmp calls `synIDattr()`. It can be solved by setting 'termguicolors' before loading nvim-cmp.

Oops, I tested this incorrectly. This doesn't work.

zeertzjq:
The following change to nvim-cmp should fix this problem:
```diff
diff --git a/plugin/cmp.lua b/plugin/cmp.lua
index 0e5e691..611b5c9 100644
--- a/plugin/cmp.lua
+++ b/plugin/cmp.lua
@@ -25,7 +25,7 @@ for kind in pairs(types.lsp.CompletionItemKind) do
   end
 end
 
-autocmd.subscribe('ColorScheme', function()
+autocmd.subscribe({ 'ColorScheme', 'UIEnter' }, function()
   highlight.inherit('CmpItemAbbrDefault', 'Pmenu', { bg = 'NONE', default = false })
   highlight.inherit('CmpItemAbbrDeprecatedDefault', 'Comment', { bg = 'NONE', default = false })
   highlight.inherit('CmpItemAbbrMatchDefault', 'Pmenu', { bg = 'NONE', default = false })
```

hrsh7th:
It was difficult to arrive at this on my own...
Thank you very much!

PR #2165, opened 2025-04-28
fix: remove redundant UIEnter subscription

crackcomm:
The UIEnter event is not needed for setting up the highlight groups in cmp.lua.

This caused a bug displaying ANSI escape sequences in neovim in tmux, kitty, and alacritty.

I found this bug after 2 days of debugging, searching for the cause in the wrong place - zsh, tmux, kitty, then testing on alacritty and xterm, and finally realizing that the issue was in the Neovim configuration itself.

If this does not in fact fix the issue, even though I tested it thoroughly, it is still a redundant call in my opinion, especially since `ColorScheme` event is immediately triggered after the subscription.

### DIFF
--- a/plugin/cmp.lua
+++ b/plugin/cmp.lua
@@ -25,7 +25,7 @@ for kind in pairs(types.lsp.CompletionItemKind) do
   end
 end
 
-autocmd.subscribe({ 'ColorScheme', 'UIEnter' }, function()
+autocmd.subscribe({ 'ColorScheme' }, function()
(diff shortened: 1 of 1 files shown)


crackcomm:
I notice that it is a fix for #1649, maybe it's appropriate to keep this functionality for `vim.fn.has('headless') == 1` - I'm not sure.

wsdjeg:
neovim supports `has('headless')`?

maybe you can check `v:argv`, use `if vim.tbl_contains(vim.v.argv, '--headless') then end`

PR #2187, opened 2025-08-28
locality: fix perf issue with long lines

### LINKED ISSUE #1841: Slowdown with large files due to locality map
### FAQ

### Announcement

This is your default suggested repro as this reproduces in the default config. That said, this also reproduces with an even *more* minimal config where the only completion source is `path`. My original repro was https://github.com/dyfrgi/slowcmp/blob/main/slowcmp.lua (based on the config from LazyVim) but I included your preferred vimscript init here as well.

```vim
if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
  },

    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}
EOF

lua << EOF

  capabilities = capabilities,
}
EOF
```

### Description

When editing a file with very long lines, starting editing is slow. This appears to be due to regexes running inside `compare.locality`, based on my profiling.

### Steps to reproduce

Open a file with a very long line, e.g. https://github.com/dyfrgi/slowcmp/blob/main/slow-lhs.ini.
Start inserting

### Expected behavior

It should start editing promptly

### Actual behavior

It's slow, taking about 15 seconds to enter insert mode

### Additional context

I think this also affects quit time.

I wasn't able to get a very good profile of this - for some reason these cmp functions don't show up in the built-in `:profile` logs, and building neovim with luajit compiled with `-DLUAJIT_USE_PERFTOOLS` still only gets me limited information. Maybe I'm using `:profile` wrong - if there's a good way to get this, I would love to know what it is. I tried starting nvim without opening the file, then running `:profile start prof.log` `:profile file *` `:profile func *`, opening the file with `:e slow-lhs.ini`, and then closing the file. Unfortunately it always winds up showing almost nothing from nvim-cmp in the log.

That said, I have some informal samples of stacks from hitting ctrl-c when noticing it was slow, and they were always this stack:
```
E5108: Error executing lua ...ocal/share/nvim/lazy/nvim-cmp/lua/cmp/config/compare.lua:161: couldn't parse regex: Keyboard interrupt
stack traceback:
	[C]: in function 'regex'
	...ocal/share/nvim/lazy/nvim-cmp/lua/cmp/config/compare.lua:161: in function 'callback'
	.../.local/share/nvim/lazy/nvim-cmp/lua/cmp/utils/cache.lua:38: in function 'ensure'
	...ocal/share/nvim/lazy/nvim-cmp/lua/cmp/config/compare.lua:159: in function 'update'
	...tenburg/.local/share/nvim/lazy/nvim-cmp/lua/cmp/init.lua:327: in function <...tenburg/.local/share/nvim/lazy/nvim-cmp/lua/cmp/init.lua:324>
	...ocal/share/nvim/lazy/nvim-cmp/lua/cmp/utils/feedkeys.lua:47: in function <...ocal/share/nvim/lazy/nvim-cmp/lua/cmp/utils/feedkeys.lua:45>
```

I'm not sure what is getting fed into it, since it seems a little odd for it to always pop out on the compile of the regex rather than the match. I have about 10 samples from hitting ctrl-c, which should be enough to see other stacks if they're common. When looking at the output from `perf`, I see about 50% of the time spent on `__memcpy_avx_unaligned_erms`, and then actually inside nvim I see a bunch of time spent on `regex_match_line` and `vim_regexec_multi;bt_regexec_both;regtry;regrepeat;utfc_ptr2len`. It's 9.3% at the top of that and 8% at the bottom, so it's spending a ton of time just iterating through the string.

At any rate, it's slow, and I don't know exactly why because I'm a neovim profiling noob.

By way of comparison, if I use a config which omits `locality` from the list of sorting comparators such as https://github.com/dyfrgi/slowcmp/blob/main/cmp-fast.vim, the issue no longer reproduces.

emmanueltouzery:
fixes #1841

i also attempted to use vim.fn.matchstrpos to avoid the substring operations, but it seems it's not really the substring that's slow, but truly the regexes (or the populating of the locality map).

this is a very simple fix that solves a really problematic issue.

### DIFF
--- a/lua/cmp/config/compare.lua
+++ b/lua/cmp/config/compare.lua
@@ -159,7 +159,8 @@ compare.locality = setmetatable({
       local locality_map = self.lines_cache:ensure({ 'line', buffer }, function()
         local locality_map = {}
         local regexp = vim.regex(config.completion.keyword_pattern)
-        while buffer ~= '' do
+        -- the buffer length check is to avoid performance issues on very long lines, #1841
+        while buffer ~= '' and #buffer < 5000 do
           local s, e = regexp:match_str(buffer)
           if s and e then
             local w = string.sub(buffer, s + 1, e)


emmanueltouzery:
note that the integration test failure is unrelated to this PR.

AckslD:
Thanks @emmanueltouzery, I was just debugging my setup and why it was slow on large single-line json files. Narrowed it down to `nvim-cmp` and the `compare.locality.update` call. Trying out your branch avoids my session freezing upon entering insert mode or the command line with such files :+1:

PR #2190, opened 2025-09-11
feat: icon and icon highlight separation

### LINKED ISSUE #1599: Changing highlight color for the icon only
Hello,
Is there a way to change the color for the icon without changing the `abbr` color as well?
Like this:
image
Asking after seeing how VSCode does it and it looked nice:
image
Thanks in advance!

tostrstrudl:
feat: allow separate highlight for completion item icon (#1599)

This PR adds support for separately highlighting the completion item icon (kind)
without affecting the abbreviation (abbr), as requested in #1599.
- Added another field ('icon') to allow for separate icon movement and highlighting
- Added new highlight groups for the icons (kind specific, icon field specific, and default)
- Added one require() line for LspKind and a 4-line function to pull a kind's icon (get_icon()); they default to cmp_opts.icon or ' ' if LspKind is not present or there are no custom-defined icons
- Updated cmp.txt to inform of the new highlight fields with an example on how to set them
- Tested with my own local configuration of nvim-cmp/lspkind to be sure that everything loads and runs as it should (no breaking changes)

Closes #1599.

### DIFF
--- a/lua/cmp/entry.lua
+++ b/lua/cmp/entry.lua
@@ -6,11 +6,15 @@ local snippet = require('cmp.utils.snippet')
 local config = require('cmp.config')
 local types = require('cmp.types')
 local matcher = require('cmp.matcher')
-local lspkind = require('lspkind') or ''
+local ok, lspkind = pcall(require, 'lspkind')
 
 local function get_icon(kind)
+  if ok then
     local icon = lspkind.symbol_map[kind]
-    return icon or ""
+    return icon
+  end
+
+  return ''
 end
 
 ---@class cmp.Entry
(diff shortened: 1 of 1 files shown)


lua/cmp/entry.lua
```diff
@@ -6,6 +6,12 @@ local snippet = require('cmp.utils.snippet')
 local config = require('cmp.config')
 local types = require('cmp.types')
 local matcher = require('cmp.matcher')
+local lspkind = require('lspkind') or ''
```
wsdjeg: maybe this should be:

```lua
local ok, lspkind = pcall(require, 'lspkind')
if ok then
    --- do something
end
```

tostrstrudl:
Everything seems to work just fine for me, but I can take a look at it over the next couple of days. I'm in between residences so I'm a bit busy packing and moving atm. Please bear with me

SteveCharleston:
No problems, in the meantime I figured it out with just slight modifications. This restores the previous display behavior:

```
                formatting = {
                    fields = { "icon", "abbr", "menu", "kind" },
                    format = function(entry, vim_item)
                        local lspkind = require("lspkind")
                        local kind = lspkind.cmp_format({ mode = "symbol_text", maxwidth = 50 })(entry, vim_item)
                        kind.icon = " " .. (kind.icon or "") .. "  "
                        kind.kind = "   (" .. (kind.kind or "") .. ")"

                        return kind
                    end,
                },
```

This might have partly been caused by https://github.com/onsails/lspkind.nvim/pull/92/changes/f8aa7dfb940ac7db1f0d4eccef53772ff91958bc

tostrstrudl:
Yeah, I think it might be. When I first did the fork I was using LspKind, but I rewrote a lot of my code for Neovim (moved to custom-ish plugins) and stopped using LspKind. I just handle the formatting on my own, so I didn't catch the change. Thank you for uploading your fix!

PR #2198, opened 2025-11-04
ci: fix `Setup tools` step failure

yuys13:
This resolves the "Setup tools" failure by upgrading gh-actions-luarocks
from v4 to v6.

This update bumps the underlying Luarocks version to 3.12.2,
which contains the fix for the "main function has more than 65536 constants"
error.

See: luarocks/luarocks#1797


hrsh7th:
Thank you. I'm glad that the test CI is back up and running.

PR #2202, opened 2025-11-13
feat: max_height for completion window

dvimaaniye:
- Max height can be set separately from ['pumheight'](https://neovim.io/doc/user/options.html#'pumheight')
- Default behaviour is still to use pumheight, so existing configs won't be affected. 
- A negative `max_height` will not display the menu. (Should it fallback to use all available space?)
```lua
	cmp.setup({
		...
		window = {
			completion = cmp.config.window.bordered({ max_height = 5 }), -- ignores pumheight
			-- completion = cmp.config.window.bordered({ max_height = 0 }), -- Uses all available space, ignores pumheight 
			-- completion = cmp.config.window.bordered(), -- Uses pumheight, default
			-- completion = cmp.config.window.bordered({ max_height = nil }), -- Uses pumheight, same as default
			
			-- completion = cmp.config.window.bordered({ max_height = -5 }), -- Menu not displayed
		},
		...
	})
```

Screenshots:
(`max_height` set to 5):
image

(`max_height` set to 0):
image

(`max_height` set to 5, `pumheight` set to 1, ignores `pumheight`):
image

(`max_height` set to `nil` (default), `pumheight` set to 1, uses `pumheight`):
image

### DIFF
--- a/doc/cmp.txt
+++ b/doc/cmp.txt
@@ -787,6 +787,13 @@ window.completion.scrollbar~
   `boolean`
   Whether the scrollbar should be enabled if there are more items that fit
 
+                                     *cmp-config.window.completion.max_height*
+window.completion.max_height~
+  `number | nil`
+  The completion window's max height, can be set to 0 to use all available
+  space. To use `vim.o.pumheight` set this to `nil`.
+  See |'pumheight'|.
+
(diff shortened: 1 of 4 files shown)


wsdjeg:
Thanks, LGTM, merged.

PR #2208, opened 2025-12-05
refactor: move shared fields into parent class

danielrainer:


danielrainer:
I just realized that the fields differ in nil-ability, so this is not a correct transform without other refactoring.

PR #2217, opened 2026-02-26
Convert vim.NIL in client response item and itemDefaults to nil

### LINKED ISSUE #2212: lua/cmp/entry.lua:351: attempt to concatenate field 'descriptio n' (a userdata value)
hey
im using ``glsl_analyzer`` and i get the error
lua/cmp/entry.lua:351: attempt to concatenate field 'description' (a userdata value)
when changing anything in the file

location of the error:
https://github.com/hrsh7th/nvim-cmp/blob/85bbfad83f804f11688d1ab9486b459e699292d6/lua/cmp/entry.lua#L351

my guess is that its caused by
``completion_item.labelDetails.description``
being ``vim.NIL`` what bypasses the if check
but i can't confirm this ;-;

my configs are the default ``nvim-cmp`` and ``lspconfig`` configs
i don't think that's the cause

ErikReider:
Fixes: #2212

### DIFF
--- a/lua/cmp/source.lua
+++ b/lua/cmp/source.lua
@@ -360,9 +360,13 @@ source.complete = function(self, ctx, callback)
         local old_offset = self.offset
         local old_entries = self.entries
 
+        response.itemDefaults = misc.ensure_nil(response.itemDefaults)
+
         self.status = source.SourceStatus.COMPLETED
         self.entries = {}
         for _, item in ipairs(response.items or response) do
+          item = misc.ensure_nil(item)
+
           if item.label then
             local e = entry.new(ctx, self, item, response.itemDefaults)
             if not e:is_invalid() then
--- a/lua/cmp/utils/misc.lua
+++ b/lua/cmp/utils/misc.lua
@@ -120,6 +120,31 @@ misc.merge = function(tbl1, tbl2)
   end
 end
 
+---Replaces all vim.NIL values with nil recursively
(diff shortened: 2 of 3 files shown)


wsdjeg:
merged, thanks:)

saatom:
Maybe it's something with my setup but this commit causes cmp to intermittently break. If I repeatedly run `nvim test.txt`, one out of every three or four times I will get the following error whenever I try to type a command in the (nvim) command line (or do anything that would normally invoke cmp). 

Reverting to commit da88697 fixes this issue, thus I suspect something in this commit is the culprit. Can anyone else corroborate this issue?
Screenshot_20260328_100614

wongak:
> Maybe it's something with my setup but this commit causes cmp to intermittently break. If I repeatedly run `nvim test.txt`, one out of every three or four times I will get the following error whenever I try to type a command in the (nvim) command line (or do anything that would normally invoke cmp).
> 
> Reverting to commit [da88697](https://github.com/hrsh7th/nvim-cmp/commit/da88697d7f45d16852c6b2769dc52387d1ddc45f) fixes this issue, thus I suspect something in this commit is the culprit. Can anyone else corroborate this issue?

@saatom I have the exact same issue and "solved" it the same way. Still trying to figure out what might cause this. I'm using lazy, so maybe there is an issue with loading the utils. But it seems really weird.

PR #2219, opened 2026-03-17
fix(#1303): TextChanged should update completion only on text added

### LINKED ISSUE #1303: Completion is triggered on text removal on the next line
### FAQ

### Announcement

```vim

if has('vim_starting')
  set encoding=utf-8
endif
scriptencoding utf-8

if &compatible
  set nocompatible
endif

end

call plug#begin(s:plug_dir)
Plug 'hrsh7th/nvim-cmp'
Plug 'hrsh7th/cmp-buffer'
Plug 'hrsh7th/cmp-nvim-lsp'
Plug 'hrsh7th/vim-vsnip'
Plug 'neovim/nvim-lspconfig'
call plug#end()
PlugInstall | quit

lua << EOF
local cmp = require "cmp"
local types = require "cmp.types"
cmp.setup {
  snippet = {
    expand = function(args)
    end,
  },

  mapping = {
    ['<Up>'] = cmp.mapping.select_prev_item({ behavior = types.cmp.SelectBehavior.Select }),
    ['<Down>'] = cmp.mapping.select_next_item({ behavior = types.cmp.SelectBehavior.Select }),
  },

    { name = "nvim_lsp" },
    { name = "buffer" },
  }),
}
EOF

lua << EOF

  capabilities = capabilities,
}
EOF

```

### Description

A corner case. When one removes the next line in the insert mode the completion menu is triggered at the end of the previous line. 
The argument against this behavior is that the user has removed the whole next line and the previous one should not be affected, hence no menu should be triggered. The current behavior breaks the insert mode navigation when the arrows & <CR> are used to also navigate the completion causing accidental completions and/or inability to use arrows to navigate the buffer.

Here's the demo:

[asciicast](https://asciinema.org/a/5p8dWftjawQ7aPjnpdurntPiU)

### Steps to reproduce

The file from the demo:

```
#!/usr/bin/env python3

import matplotlib.pyplot as plt

def nxtpwr(t):
    if t < 8:
        return 8

    t -= 1
    t |= t >> 1
    t |= t >> 2
    t |= t >> 4
    t |= t >> 8
    t |= t >> 16
    t += 1

    return t

plt.plot([nxtpwr(x) for x in range(1, 100)])
plt.show()
```

1. Navigate to the empty line
2. Enter the insert mode
3. Type in some text
5. Press backspace until the whole line is removed and the cursor goes to the previous line end

### Expected behavior

I would expect the completion NOT to be triggered.

### Actual behavior

The completion is triggered.

### Additional context

While it may be uncommon, I've got used to navigating & removing text in the insert mode from time to time (also `vim-rsi` contributes to this habit), and having the menu at the end of the previous line is locking me in the menu.

Before nvim-cmp I'd been using CoC.nvim and it was handling such corner cases somehow. Even if the resolution will be no-fix, a clue on how to handle this case in the config would be very appreciated.

litoj:
ensure that the text change we react to is only when a character is written.

Resolves #1303


wsdjeg:
merged, thanks.

eeeXun:
Hi, after this changed, the `nvim-cmp` behaves oddly. I'm sure whether it's a feature or a bug.

If I type some text first. It then shows the completion menu. But when I delete the text, the completion menu stays open and show all the unrelated completions what I typed before. I'd expect the menu to close, since the current text has been cleared.

Gently pin @wsdjeg to take a look

litoj:
Well, it should update when you write anything after the deletion, right?

I suppose if you delete more than a word then the completion does become irrelevant. That could be improved, yes.

My aim with this was to avoid the flicker that happened whenever I fixed a typo. And when you actually want the completion, it will still be up-to-date. Just not while you are deleting text.

eeeXun:
> Well, it should update when you write anything after the deletion, right?
> 
> I suppose if you delete more than a word then the completion does become irrelevant. That could be improved, yes.
> 
> My aim with this was to avoid the flicker that happened whenever I fixed a typo. And when you actually want the completion, it will still be up-to-date. Just not while you are deleting text.

Okey, I get your point. But this is the problem that I'm facing 

https://github.com/user-attachments/assets/bd28fd0d-7766-4ca3-a432-ec32f885788e

I deleted all the text, but the menu is still open.

Also some problem with snippet. With following example

```snippets
snippet test
	Snippet test
	${0}
```

https://github.com/user-attachments/assets/71c1c6a7-3806-44be-9c21-d0a023a816a4

After accepting the snippet, the menu remains open. The menu should close.

